PR #2, opened 2019-03-12
Add Mobius.swift logo

adamhowardprice:


togi:
I'm pretty sure using the Swift logo like this is against the guidelines that Apple provide, very much like using a Spotify logo in the same way would be against our own guidelines.

Swift logo guidelines can be found here: https://developer.apple.com/swift/resources/

For the time being you could use the logo that's in the wiki that doesn't include the swift logotype

PR #5, opened 2019-03-12
Fix code coverage

dflems:
* Build to a different derived data path for test as codecov was picking up the wrong files
* Tweak the config a little


dflems:
Codecov seems to recognize that I _want_ to ignore certain paths but isn't actually doing it...

dflems:
It may only update once merged to master

PR #7, opened 2019-03-13
Add code samples to the readme

jeppes:
The readme now provides a high-level overview of the core concepts: `Models`, `Events`, and `Effects`

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,10 +20,99 @@ This repository contains the core Mobius framework and add-ons for common develo
 | 🛠 Xcode    | 10.1+       |
 | 🐦 Language | Swift 4.2  |
 
-## Usage
+## Usage in your project
 Pull in Mobius.swift as a dependency. Either as a submodule or using [Carthage](https://github.com/Carthage/Carthage).
 
 Build the project and link with the frameworks.
+
+## Mobius in action - Building a Counter
+
+The goal of Mobius is to you better control over your application state. You can think of your state as a snapshot of all the current values of the variables in your application.In Mobius, we take encapsulate all of the state in a data-structure which we call the *Model*.
+
+The *Model* can be represented by whatever type you like. Since we're building a counter, we'll be able to encapsulate all of our state in an `Int`:
+
+```swift
+typealias MyModel = Int
+```
+
+Mobius does not let you manipulate the state directly. In order to change the state, you have to send the framework messages saying what you want to do. We call these messages *Events*. In our case, we'll want to be able to increment and decrement our counter. Let's use an `enum` to define these cases:
+```swift
+enum MyEvent {
+    case increment
+    case decrement
+}
+```
+
+Now that we have a *Model* and some *Event*s, we'll need to give Mobius a set of rules which it can use to update the state on our behalf. We do this by giving the framework a function which it can call with its latest *Model* and any incomming *Event*, in order to generate a new *Model*:
+```swift
+func update(model: MyModel, event: MyEvent) -> MyModel {
+    switch event {
+    case .increment: return model + 1
+    case .decrement: return model - 1
+    }
+}
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,11 +20,100 @@ This repository contains the core Mobius framework and add-ons for common develo
...
 Build the project and link with the frameworks.
 
+## Mobius in action - Building a Counter
+
+The goal of Mobius is to you better control over your application state. You can think of your state as a snapshot of all the current values of the variables in your application.In Mobius, we take encapsulate all of the state in a data-structure which we call the *Model*.
+
+The *Model* can be represented by whatever type you like. Since we're building a counter, we'll be able to encapsulate all of our state in an `Int`:
+
+```swift
+typealias MyModel = Int
```
BalestraPatrick: Can we rename these entities to something like `CounterModel` and `CounterEvent` instead of `My`?

README.md
```diff
@@ -20,11 +20,100 @@ This repository contains the core Mobius framework and add-ons for common develo
...
 
-## Usage
+## Usage in your project
 Pull in Mobius.swift as a dependency. Either as a submodule or using [Carthage](https://github.com/Carthage/Carthage).
 
 Build the project and link with the frameworks.
 
+## Mobius in action - Building a Counter
+
+The goal of Mobius is to you better control over your application state. You can think of your state as a snapshot of all the current values of the variables in your application.In Mobius, we take encapsulate all of the state in a data-structure which we call the *Model*.
```
BalestraPatrick: missing `give`?

README.md
```diff
@@ -20,11 +20,102 @@ This repository contains the core Mobius framework and add-ons for common develo
...
 
-## Usage
+## Usage in Your Project
 Pull in Mobius.swift as a dependency. Either as a submodule or using [Carthage](https://github.com/Carthage/Carthage).
 
 Build the project and link with the frameworks.
 
+## Mobius in Action - Building a Counter
+
+The goal of Mobius is to give you better control over your application state. You can think of your state as a snapshot of all the current values of the variables in your application. In Mobius, we take encapsulate all of the state in a data-structure which we call the *Model*.
```
togi: "we take encapsulate"

README.md
```diff
@@ -20,11 +20,102 @@ This repository contains the core Mobius framework and add-ons for common develo
...
+
+The *Model* can be represented by whatever type you like. Since we're building a counter, we'll be able to encapsulate all of our state in an `Int`:
```
togi: suggestion: "In this example  we'll be building a simple counter"

README.md
```diff
@@ -20,11 +20,102 @@ This repository contains the core Mobius framework and add-ons for common develo
...
+
+Mobius does not let you manipulate the state directly. In order to change the state, you have to send the framework messages saying what you want to do. We call these messages *Events*. In our case, we'll want to be able to increment and decrement our counter. Let's use an `enum` to define these cases:
+```swift
+enum CounterEvent {
+    case increment
+    case decrement
+}
+```
+
+Now that we have a *Model* and some *Event*s, we'll need to give Mobius a set of rules which it can use to update the state on our behalf. We do this by giving the framework a function which it can call with its latest *Model* and any incomming *Event*, in order to generate a new *Model*:
```
togi: "will be sequentially called with every incoming *Event* and the most recent *Model*, in order to generate the next *Model*"

README.md
```diff
@@ -20,11 +20,102 @@ This repository contains the core Mobius framework and add-ons for common develo
...
+```swift
+func update(model: CounterModel, event: CounterEvent) -> CounterModel {
+    switch event {
+    case .increment: return model + 1
+    case .decrement: return model - 1
+    }
+}
+```
+
+With these building blocks, we can start to think about our applications as transitions between discrete states in response to events. But we believe there is one piece still missing from the puzzle - namely the side-effects which are associated with moving between states. For instance, pressing a "refresh button" might put our application into a "loading" state, with the side-effect of also fetching the latest data from our backend.
```
togi: "refresh" button

PR #9, opened 2019-03-13
Add initial podspecs

dflems:
Tested against a local specs repo. Spec lint checks passed. Shipped as four separate podspecs as you seemingly cannot have subspecs define their own module.

I'm also working on a script to generate these files and deploy them to the cocoapods trunk repo, which I can only really test once this is merged. When I get the script to a point where it works, I will submit it in a followup PR, along with a README section that explains how to integrate with cocoapods.

@BalestraPatrick @jeppes

### DIFF
--- a/MobiusCore.podspec.json
+++ b/MobiusCore.podspec.json
@@ -1,0 +1,22 @@
+{
+  "name": "MobiusCore",
+  "version": "0.1-alpha",
+  "summary": "A functional reactive framework for managing state evolution and side-effects",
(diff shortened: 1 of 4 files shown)


dflems:
Published!

* https://cocoapods.org/pods/MobiusCore
* https://cocoapods.org/pods/MobiusExtras
* https://cocoapods.org/pods/MobiusTest
* https://cocoapods.org/pods/MobiusNimble

PR #13, opened 2019-03-14
Make private Cartfile public in order to fix installation via Carthage

BalestraPatrick:
It seems like that specifying `Quick` and `Nimble` in the public Cartfile is the only way to make Carthage happy when building `MobiusNimble`. Since Carthage doesn't support subspecs, client will need to pull down test dependencies even though they may not be interested in using them together with `MobiusNimble`. They can obviously always choose to integrate the frameworks into their project or not.
This is what has been suggested by other OSS projects too such as [Moya](https://github.com/Moya/Moya#carthage).


dflems:
I think you just need `Nimble`. We only use `Quick` in test targets.

dflems:
Either way I don't think it'll hurt!

BalestraPatrick:
@dflems You're right, only moved `Nimble` to the public `Cartfile`.

PR #23, opened 2019-04-11
Move umbrella scheme to different directory and dynamically use it only on CI

BalestraPatrick:
This will improve performance when integrating the library via Carthage. Right now, Carthage takes every xcscheme and builds it and there's only a few hacky ways to get around it. This seems to be best solution at this moment.

@rastersize


rastersize:
Thanks @BalestraPatrick!

PR #28, opened 2019-05-27
Add CompositeEventSourceBuilder

JensAyton:
We currently have `MergedEventSource` to compose multiple event sources with the same event type, but it’s unergonomic because it takes an array of event sources, but you can’t construct an array of heterogeneous event sources due to limitations on protocols with associated types.

The current workaround is to explicitly wrap the members of the array in `AnyEventSource`, but explicit use of type erasure in clients is unpleasant.

Using a builder lets us take each individual event source as a generic parameter and handle the type erasure internally.

I called it `CompositeEventSourceBuilder` rather than `MergedEventSource` by analogy to `CompositeDisposable`.

### DIFF
--- a/MobiusCore/Source/EventSources/CompositeEventSourceBuilder.swift
+++ b/MobiusCore/Source/EventSources/CompositeEventSourceBuilder.swift
@@ -1,0 +1,62 @@
+// Copyright (c) 2019 Spotify AB.
+//
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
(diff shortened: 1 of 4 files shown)


JensAyton:
Draft – missing unit tests

MobiusCore/Source/EventSources/CompositeEventSourceBuilder.swift
```diff
@@ -0,0 +1,62 @@
...
+    -> CompositeEventSourceBuilder<Event> where Source.Event == Event {
+        let es = AnyEventSource<Event>(source)
+        let sources = eventSources + [es]
+        return CompositeEventSourceBuilder(eventSources: sources)
+    }
+
+    /// Builds an event source that composes all the event sources that have been added to the builder.
+    ///
+    /// - Returns: An event source which represents the composition of the builder’s input event sources. The type
+    /// of this source is an implementation detail; consumers should avoid spelling it out if possible.
```
JensAyton: In particular, I’d like the return type to be `some EventSource<.Event == Event>`, but Swift 5.1 will only have unconstrained opaque types.

JensAyton:
Now with unit tests.

In order to let the existing `MergedEventSource` tests run with `MergedEventSource` deprecated, I had to disable `-warnings-as-errors` in the test target. Unfortunately [there is no finer-grained control for deprecation warnings in Swift](https://forums.swift.org/t/swift-should-allow-for-suppression-of-warnings-especially-those-that-come-from-objective-c/19216/5).

PR #29, opened 2019-06-04
Support Swift 5.1 toolchains

JensAyton:
In older toolchains, `==` must have a dummy return statement. In newer ones, it must not. (Having one generates an unreachable code warning, and we treat warnings as errors.)

I’d like to formally deprecate `NoEffect` in favour of `Never`. Unfortunately, there is no way to make an availability attribute conditional on the version of Swift used by the _consumer_ of the library, so as long as we support Swift 4.x (where `Never` isn’t `Hashable` or `Equatable`) we can’t do this.

If such a deprecation isn’t controversial, I guess we can document it as soft-deprecated.

### DIFF
--- a/MobiusCore/Source/NoEffect.swift
+++ b/MobiusCore/Source/NoEffect.swift
@@ -26,7 +26,14 @@ import Foundation
 ///         conformance is just to allow the type to be used where required. Such as with the `First` and `Next` data
 ///         structures.
 public enum NoEffect: Equatable, Hashable {
-    public static func == (lhs: NoEffect, rhs: NoEffect) -> Bool { return true }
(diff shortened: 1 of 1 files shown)


dflems:
Gonna merge this. The rule that coverage can never ever ever be decreased (even by 0.68%) is a bit egregious.

dflems:
cc @jeppes

JensAyton:
More importantly, it didn’t add any actual code and the code that is there can never be executed, so it’s quite hard to cover.

dflems:
I think it's that `NoEffect.swift` had no coverage before but now it has more lines of code (it counts the line with the `#if` and `#else`).

So coverage did go down, just barely.

PR #31, opened 2019-06-14
 Use full names for generic type arguments

JensAyton:
Type parameters like `<T: LoopTypes>` where `T` is constrained and has semantics are unidiomatic in Swift. This PR cleans up all the cases on parameters of generic types (there are still many function parameters like `addThing<T: Thing>()`).

### DIFF
--- a/MobiusCore/Source/Next.swift
+++ b/MobiusCore/Source/Next.swift
@@ -72,13 +72,11 @@ public extension Next {
     var hasEffects: Bool { return !effects.isEmpty }
 }
 
-#if swift(>=4.1)
 extension Next: Equatable where Model: Equatable {
     public static func == (lhs: Next<Model, Effect>, rhs: Next<Model, Effect>) -> Bool {
         return lhs.model == rhs.model && lhs.effects == rhs.effects
     }
 }
-#endif
 
 extension Next: CustomDebugStringConvertible {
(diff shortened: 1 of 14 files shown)


MobiusExtras/Source/EventSource+Extensions.swift
```diff
@@ -26,7 +26,7 @@ public extension EventSource {
     /// - Parameters:
     ///   - map: Translation function to apply to the forwarded events.
     /// - Returns: An `EventSource` that translates and forwards event from the receiver.
-    func map<U>(_ map: @escaping (Event) -> U) -> AnyEventSource<U> {
+    func map<T>(_ map: @escaping (Event) -> T) -> AnyEventSource<T> {
```
JensAyton: This is a special case; `T` is used as the return type in standard library map functions.

MobiusCore/Source/Next.swift
```diff
@@ -72,13 +72,11 @@ public extension Next {
     var hasEffects: Bool { return !effects.isEmpty }
 }
 
-#if swift(>=4.1)
```
JensAyton: Obsolete, we require Swift 4.2.

rastersize:
@JensAyton I say let’s merge this and disregard the code coverage “failure”.

PR #35, opened 2019-08-09
Upgrade to Swift 5 and apply Xcode fixes

BalestraPatrick:
Mobius.swift is using a pretty old version of Quick and Nimble which is causing SPM to not resolve the dependency graph in some cases (for example when used to together with [XCLogParser](https://github.com/spotify/xclogparser/) which uses Commandant which itself has a dependency on Quick 2.0.0 and Nimble 8.0.0). For this reason, it's good to move to the newer versions. 

~I took the opportunity to also update to Swift 5 and apply all Xcode 10.2.1 suggested fixes in 
4d5d5bc.~ 
Update: Turns out these changes are not needed at all, it probably was a temporary Xcode glitch.

~Note: the new version of Quick that we have updated to has some warnings that are still unresolved (https://github.com/Quick/Nimble/pull/661).~ 
Update: These warnings don't show up anymore on my machine for some reason, even better.


Tools/cocoapods.rb
```diff
@@ -20,7 +20,7 @@
 end
 
 # Common Settings
-SWIFT_VERSION = '4.2'
+SWIFT_VERSION = '5.0'
```
manucheri: We should update the README.md with the new Swift (and Xcode?) requirements as well.

BalestraPatrick: We also probably want to keep a Swift 4.2 branch and `Package@swift-4.2.swift`. I'll push a new commit.

BalestraPatrick: There were no source changes necessary for migrating to Swift 5, so I think having a `Package@swift-4.2.swift` should suffice.

Package.swift
```diff
@@ -11,8 +11,8 @@ let package = Package(
         .library(name: "MobiusTest", targets: ["MobiusTest"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/Quick/Nimble", .upToNextMajor(from: "7.0.0")),
-        .package(url: "https://github.com/Quick/Quick", .upToNextMinor(from: "1.2.0")),
+        .package(url: "https://github.com/Quick/Nimble", from: "8.0.0"),
+        .package(url: "https://github.com/Quick/Quick", from: "2.0.0"),
```
JensAyton: I believe this should be `from: "2.1.0"`, since Carthage users are free to start depending on hypothetical 2.1 features

Package.swift
```diff
@@ -25,5 +25,5 @@ let package = Package(
         .testTarget(name: "MobiusNimbleTests", dependencies: ["MobiusNimble", "Quick"], path: "MobiusNimble/Test"),
         .testTarget(name: "MobiusTestTests", dependencies: ["MobiusTest", "Quick"], path: "MobiusTest/Test"),
     ],
-    swiftLanguageVersions: [.v4_2]
+    swiftLanguageVersions: [.v5]
```
JensAyton: `[.v4_2, .v5]` for consistency?

BalestraPatrick: Why is that? This `Package.swift` will be used when targeting Swift 5, and `Package@swift-4.2.swift` when targeting Swift 4.2

Package.swift
BalestraPatrick: Updated

Package.swift
JensAyton: Bad brain moment there

PR #36, opened 2019-08-28
Add an Event Source Convenience Class

laptobbe:
## Background
Since mobius does not guarantee what thread the consumer of an event source will be called on. Or what thread subscribe or dispose will be called on. The standard practice is to implement locking for every event source. When the number of event sources grows so does the code duplication.

Another problem is that the lock implementation takes away focus from actual implementation of the event source. Making the code harder to understand, especially for those new to mobius.

## Solution
One solution, that I have implemented in this PR, is to create a convenience super class to contain the locking logic. Similar to the ConnectableClass for connectables. This implementation will not work for every case, which is why my suggestion is to put it in the mobius extras.

## NSObject superclass
The event source class subclasses NSObject, I did this to avoid having to have nested or private classes that subclasses NSObject. My thinking is that event sources will interact with the outside world, and on iOS, Mac that means a NSObject subclass to be able to implement a delegate protocol for example. This also comes from having to create private classes (a lot) when using ConnectableClass. If this is not what we want, we can easily change that.


jeppes:
Sorry about the slow feedback time on this one. Addressing the issues with event sources is in our backlog, but it's unfortunately not something we are able to prioritise at this time.

PR #37, opened 2019-09-03
Adds a ‘CompositeLogger’ and builder that joins multiple loggers toge…

joelevin:
…ther

This brings some feature parity between this iOS implementation and the Android one.
I essentially copied the Android implementation and added a builder to do the type erasure `AnyMobiusLogger` (similar to `AnyEventSource`)

c.f. https://github.com/spotify/mobius/blob/master/mobius-extras/src/main/java/com/spotify/mobius/extras/CompositeLogger.java


pettermahlen:
Should this be in MobiusExtras?

joelevin:
Yeah… not sure.  I inquired in Slack and can move it to Extras if needed.

MobiusCore/Source/CompositeLoggerBuilder.swift
```diff
@@ -0,0 +1,49 @@
...
+    /// - Returns: a new `CompositeEventSourceBuilder` with the logger added to it
+    public func addLogger<Logger: MobiusLogger>(_ logger: Logger) -> CompositeLoggerBuilder<T> where Logger.Effect == T.Effect, Logger.Model == T.Model, Logger.Event == T.Event {
+        let loggers = mergedLoggers + [AnyMobiusLogger<T>(logger)]
+        return CompositeLoggerBuilder(loggers: loggers)
+    }
+
+    /// Builds a single logger that composes all loggers that have been added to the builder
+    ///
+    /// - Returns: A MobiusLogger that forwards all logging messages to each of the builder's input loggers. The type
+    /// of this logger is an implementation detail; consumers should avoid making AnyMobiusLoggers if possible.
```
JensAyton: Is this intended to say “should avoid making `CompositeMobiusLogger`s if possible”?

This seems unnecessarily vague. We should wither make `CompositeMobiusLogger` internal, or make `build` return a `CompositeMobiusLogger` and not have notes like this.

I lean towards having `CompositeMobiusLogger` public and use of the builder optional, since builders aren’t idiomatic Swift.

MobiusCore/Source/CompositeLoggerBuilder.swift
```diff
@@ -0,0 +1,49 @@
...
+
+public struct CompositeLoggerBuilder<T: LoopTypes> {
+    private let mergedLoggers: [AnyMobiusLogger<T>]
+
+    /// Initializes a `CompositeLoggerBuilder`
+    public init() {
+        self.init(loggers: [])
+    }
+
+    private init(loggers: [AnyMobiusLogger<T>]) {
```
JensAyton: If we do want to hide/discourage `CompositeMobiusLogger` (see below), this should be a single public init with a default empty array.

PR #39, opened 2019-09-26
Add an EffectAction class

zvonicek:
This PR contains a general implementation of the `ActionWithPredicate` protocol that takes a closure and triggers it when the provided effect or precidate is accepted and the action runs.

We identified a common pattern in our codebase of repeating simple, single-purpose `Action` classes eg. to handle various navigation events. This class helped us to generalize them. 

Support for effect cases with associated values does not come out of the box because of the Swift limitations, but it can be done by providing a custom predicate. Example of such use case is in unit tests.

### DIFF
--- a/MobiusExtras/Source/EffectAction.swift
+++ b/MobiusExtras/Source/EffectAction.swift
@@ -1,0 +1,47 @@
+// Copyright (c) 2019 Spotify AB.
+//
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import MobiusCore
+
+/// A general implementation of the `ActionWithPredicate` protocol that takes an effect and a closure
+/// and triggers the closure when the effect is accepted and the action runs.
+public class EffectAction<EffectType>: ActionWithPredicate where EffectType: Equatable {
+    public typealias Effect = EffectType
+
+    private let acceptedEffect: Effect
+    private let action: () -> Void
+
+    /// Initializes an EffectAction
+    ///
+    /// - Parameters:
+    ///   - acceptedEffect: en effect that the EffectAction accepts.
+    ///   - action: a closure that gets triggered on the action run when the effect is accepted.
+    public init(_ acceptedEffect: Effect, action: @escaping () -> Void) {
+        self.acceptedEffect = acceptedEffect
+        self.action = action
+    }
+
+    public func canAccept(_ effect: Effect) -> Bool {
+        return effect == acceptedEffect
+    }
+
+    public func run() {
+        action()
+    }
+}
--- a/MobiusExtras/Source/EffectAction.swift
+++ b/MobiusExtras/Source/EffectAction.swift
@@ -20,7 +20,7 @@
 import MobiusCore
 
 /// A general implementation of the `ActionWithPredicate` protocol that takes a closure and triggers it when
-/// the provided effect or precidate is accepted and the action runs.
+/// the provided effect or predicate is accepted and the action runs.
 public class EffectAction<Effect>: ActionWithPredicate where Effect: Equatable {
     public typealias Predicate = (Effect) -> (Bool)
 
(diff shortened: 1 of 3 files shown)


MobiusExtras/Source/EffectAction.swift
```diff
@@ -0,0 +1,56 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import MobiusCore
+
+/// A general implementation of the `ActionWithPredicate` protocol that takes a closure and triggers it when
+/// the provided effect or precidate is accepted and the action runs.
```
pettermahlen:
```suggestion
/// the provided effect or predicate is accepted and the action runs.
```

MobiusExtras/Source/EffectAction.swift
```diff
@@ -0,0 +1,56 @@
...
+public class EffectAction<Effect>: ActionWithPredicate where Effect: Equatable {
+    public typealias Predicate = (Effect) -> (Bool)
+
+    private let predicate: Predicate
+    private let action: () -> Void
+
+    /// Initializes an EffectAction with a predicate.
+    ///
+    /// - Parameters:
+    ///   - predicate: a predicate whose accepting inputs the EffectAction accepts.
```
pettermahlen:
```suggestion
    ///   - predicate: a predicate selecting which Effects the EffectAction accepts.
```

MobiusExtras/Source/EffectAction.swift
```diff
@@ -0,0 +1,56 @@
...
+    ///   - action: a closure that gets triggered on the action run when the effect is accepted.
```
pettermahlen:
```suggestion
    ///   - action: a closure that gets executed when the EffectAction runs.
```

MobiusExtras/Source/EffectAction.swift
```diff
@@ -0,0 +1,56 @@
...
+    public init(predicate: @escaping Predicate, action: @escaping () -> Void) {
+        self.predicate = predicate
+        self.action = action
+    }
+
+    /// Initializes an EffectAction with an accepted effect.
+    ///
+    /// - Parameters:
+    ///   - acceptedEffect: en effect that the EffectAction accepts.
+    ///   - action: a closure that gets triggered on the action run when the effect is accepted.
```
pettermahlen:
```suggestion
    ///   - action: a closure that gets executed when the EffectAction runs.
```

MobiusExtras/Test/EffectActionTests.swift
```diff
@@ -0,0 +1,105 @@
...
+            context("when initializing with accepted effect") {
+                beforeEach {
+                    effectAction = EffectAction(.accepted, action: actionClosure)
+                    actionCalled = false
+                }
+
+                context("when asking for effect acceptance on acceptable effect") {
+                    it("is accepted") {
+                        expect(effectAction.canAccept(.accepted)).to(beTrue())
+                        expect(actionCalled).to(beFalse())
```
pettermahlen: What is the reason for validating that the action hasn't been called? That seems to be outside of the scope of the test case. Not a big deal, of course, but just a bit of extra noise in the tests.

MobiusExtras/Test/EffectActionTests.swift
```diff
@@ -0,0 +1,105 @@
...
+                    }
+                }
+
+                context("when asking for effect acceptance on non acceptable effect") {
+                    it("is not accepted") {
+                        expect(effectAction.canAccept(.rejected)).to(beFalse())
+                        expect(actionCalled).to(beFalse())
```
pettermahlen: As above.

MobiusExtras/Test/EffectActionTests.swift
```diff
@@ -0,0 +1,105 @@
...
+                            return false
+                        },
+                        action: actionClosure)
+                    actionCalled = false
+                }
+
+                context("when asking for effect acceptance on acceptable effect") {
+                    it("is accepted") {
+                        expect(effectAction.canAccept(.acceptedWith(associatedValue: false))).to(beTrue())
+                        expect(actionCalled).to(beFalse())
```
pettermahlen: as above

MobiusExtras/Test/EffectActionTests.swift
```diff
@@ -0,0 +1,105 @@
...
+                    }
+                }
+
+                context("when asking for effect acceptance on non acceptable effect") {
+                    it("is not accepted") {
+                        expect(effectAction.canAccept(.rejected)).to(beFalse())
+                        expect(actionCalled).to(beFalse())
```
pettermahlen: As above.

MobiusExtras/Test/EffectActionTests.swift
zvonicek: My original intention was to put it there as a counterweight to `beTrue()` check of it in `when running the action`, but I agree it's not really needed -- I'll remove it.

jeppes:
We are creating a single API for effect handlers which we hope to migrate to. The other methods of creating effect handlers will be deprecated. Once that is in place, I don't think this will be necessary.
@zvonicek

zvonicek:
The proposed Effect Handler simplification makes this unnecessary, so I'm closing the PR.

PR #44, opened 2019-10-15
Remove public extension on NSRecursiveLock

JensAyton:
Exporting this was a wart required by ObjC semantics, and a wrapping struct is a zero-cost abstraction.

@jeppes


jeppes:
Definitely agree, but this is technically a breaking change. Should we deprecate now and remove in 1.0 instead?

JensAyton:
This is the `t3-2019-updates`, which is expected to have breaking changes.

PR #45, opened 2019-10-16
Threading changes WIP

JensAyton:
⚠️ Do not merge – I haven’t figured out the implications for MobiusController yet

Removes all knowledge of dispatch queues from raw Mobius loops. Instead, it adds a simple single-threaded work queue to sequence events and effects.

> The ordering guarantee we want to commit to at this time is: if two events E1 and E2 are dispatched, all effects produced by E1 will be handled before any effects produced by E2.
>
> (Note that an effect is considered “handled” when the effect handler’s accept returns. The effect handler may still be doing work asynchronously.)
>
> The easiest way to provide this guarantee is to fully sequence all events and effects. Combined with the use of arrays for effects that means execution order is fully deterministic and in the obvious order, but we’re not committing to that right now.

It also adds `SequentialAccessGuard`, which is like a mutex except it deliberately explodes (in debug builds) if the critical region is accessed from multiple threads. This is used to guard against concurrent access to `MobiusLoop`s.


.swiftlint.yml
```diff
@@ -81,7 +80,7 @@ redundant_nil_coalescing: error
 trailing_comma:
   mandatory_comma: true
 vertical_whitespace:
-  max_empty_lines: 1
+  max_empty_lines: 2
```
jeppes: Can we just remove this check entirely?

JensAyton: Right here I’m just going for alignment with our internal config

JensAyton:
Even wippier wip [here](https://github.com/JensAyton/Mobius.swift/commit/d242309e513948c848e62eefcc78998309836927) (not expected to work)

JensAyton:
Updated with MobiusController set up the way I want it, although there is a failing test.

The bigger issue is that there is no way for the MobiusController to ensure that events posted from an asynchronous effect handler are marshalled back to the `loopQueue`. I don’t particularly want to expose the `loopQueue` and have effect handlers deal with it themselves, and I don’t want a private hook just for MobiusController.

PR #48, opened 2019-10-28
Simplified Effect Handlers

jeppes:
## Context
Effect handlers have a range of problems:
1. There are too many ways of creating effect handlers.
2. Effect handlers have implicit threading semantics which are often neglected or simply forgotten about when using anything besides `ConnectableClass`.
3. Unwrapping Effects with associated values requires duplication and is not entirely type-safe.

## Goal
This PR aims to consolidate the creation of EffectHandlers into a simple API which internally conforms to our threading expectations, provides a convenient way to unwrap effects without duplication, and deprecates all other ways of creating effect handlers.

## Future Steps
If this is merged, we will need to:
- [ ] Update README
- [ ] Update Wiki
- [ ] Update Playground

@JensAyton 

cc: @togi @pettermahlen @rastersize @anawara @BalestraPatrick
cc: @zvonicek This should address the issues you outlined in https://github.com/spotify/Mobius.swift/pull/39


JensAyton:
Since this introduces deprecations, it should target t3-2019-updates instead of master.

MobiusCore/Source/EffectHandlers/EffectHandler.swift
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    public func canAccept(_ effect: Effect) -> Bool {
+        return canAcceptEffect(effect)
+    }
+
+    public func connect(_ consumer: @escaping (Event) -> Void) -> Connection<Effect> {
+        lock.lock()
+        defer { lock.unlock() }
+        self.consumer = consumer
```
pettermahlen: This implicitly reconnects to a new consumer if already connected. There is nothing in the documentation about the semantics of reconnecting. In my opinion, reconnecting would most likely be done due to some error. Has that been considered?

MobiusCore/Source/EffectHandlers/EffectHandler.swift
```diff
@@ -0,0 +1,142 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+import Foundation
+
+final public class EffectHandler<Effect, Event> {
+    public typealias InputType = Effect
```
pettermahlen: I don't see these type aliases used anywhere?

MobiusCore/Source/EffectHandlers/EffectHandler.swift
```diff
@@ -0,0 +1,142 @@
...
+            switch canAccept(effect) {
+            case .handle(let subType):
+                handleEffect(subType, dispatch)
+            default:
+                fatalError("ActionEffectHandler's canAccept is implemented incorrectly. This should not be possible")
+            }
+        }
+    }
+
+    public func canAccept(_ effect: Effect) -> Bool {
```
pettermahlen: Should this be public? If so, why?

MobiusCore/Source/EffectHandlers/EffectHandler.swift
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    private func dispatch(event: Event) {
+        lock.lock()
+        defer { lock.unlock() }
+        consumer?(event)
+    }
+
+    private func dispose() {
+        // Make sure that the `Connection` produced in the `connect` function is never used again
```
pettermahlen: I guess this is a TODO note?

pettermahlen: If so, care needs to be taken to handle multiple connections returned.

MobiusCore/Source/EffectHandlers/EffectHandler.swift
jeppes: No, it should be internal

MobiusCore/Source/EffectHandlers/EffectHandler.swift
jeppes: This is from the template. I will remove it.

MobiusCore/Source/EffectHandlers/EffectHandler.swift
jeppes: You are right.
The docs for `Connectable` state that `Connectable`s should either support multiple independent connections or throw an exception. I went with the latter in this case.

jeppes:
@pettermahlen thanks for your feedback. I have addressed your comments and I would appreciate it if you could take another look.

MobiusCore/Source/EffectHandlers/EffectHandler.swift
jeppes: They are gone now.

MobiusCore/Source/EffectHandlers/EffectHandler.swift
```diff
@@ -0,0 +1,131 @@
...
+    /// Create a handler for effects which satisfy the `canAccept` parameter function.
+    ///
+    /// - Parameter canAccept: A function which indicates whether to handle an effect. If it returns `.handle(effect)` for a given `effect`, this
+    /// effect handler will handle said effect.
+    /// - Parameter handleEffect: Handle effects which satisfy `canAccept`.
+    /// - Parameter onDispose: Tear down any resources being used by this effect handler.
+    public init<AssociatedValueType>(
+        canAccept: @escaping (Effect) -> HandleEffect<AssociatedValueType>,
+        handleEffect: @escaping (AssociatedValueType, @escaping Consumer<Event>) -> Void,
+        onDispose disposable: @escaping () -> Void
```
JensAyton: The biggest point-of-use difference between our implementations is that I bundled these three functions into a protocol. Apart from a general feeling that “protocols are more Swifty”, I found that this was quite nice to use in that it felt natural to group the `canAccept` and `handleEffect` parts of each effect into a struct or final class depending on statefulness. It also led to composition like this:

```swift
func myFeatureEffectHandler(thingManager: ThingManager) -> CompositeEffectHandler<MyLoopTypes> {
    return CompositeEffectHandler()
        .addHandler(MyFirstEffectHandler())
        .addHandler(MySecondEffectHandler(thingManager: thingManager))
}
```

How will features look with the closure-based design? I imagine developers will either end up bundling effect handlers in types anyway to manage state, or creating a hierarchy of factory functions with a bunch of closure captures that may make debugging and memory management harder.

jeppes:
I have now clarified the semantics around the EffectHandler lifecycle and updated the tests to match these semantics.

MobiusExtras/Test/ConnectableClassTests.swift
```diff
@@ -1,148 +0,0 @@
...
-// specific language governing permissions and limitations
-// under the License.
-
-import MobiusCore
-import Nimble
-import Quick
-
-@testable import MobiusExtras
-
-class ConnectableTests: QuickSpec {
```
JensAyton: I’d prefer not to delete tests of deprecated stuff – if it’s available, it should be tested. In at least one of my working branches I disabled warnings-as-errors in order to keep deprecated tests around for now

MobiusCore/Source/EffectHandlers/EffectHandler.swift
```diff
@@ -0,0 +1,142 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+import Foundation
+
+final public class EffectHandler<Effect, Event> {
+    public typealias InputType = Effect
+    public typealias OutputType = Event
+
+    private let lock = NSRecursiveLock()
```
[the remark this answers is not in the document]
jeppes: Good idea!

MobiusCore/Source/EffectRouterBuilder.swift
```diff
@@ -66,39 +45,19 @@ public struct EffectRouterBuilder<Input, Output> {
...
-
-    public func addAction<A: ActionWithPredicate>(_ action: A, queue: DispatchQueue? = nil) -> EffectRouterBuilder<Input, Output> where A.Effect == Input {
-        let connectable = ClosureConnectable<Input, Output>(action.run, queue: queue)
-        return addConnectable(connectable, predicate: action.canAccept)
+    public func addEffectHandler(
+        _ effectHandler: EffectHandler<Input, Output>
+    ) -> EffectRouterBuilder<Input, Output> {
+        let handler = (connect: effectHandler.connect, predicate: effectHandler.canAccept)
+        return EffectRouterBuilder<Input, Output>(connectables: connectables + [handler])
     }
```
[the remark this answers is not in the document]
jeppes: I don't know what the larger plan is, but I personally do not want to keep `EffectRouterBuilder`.
`EffectHandler` should compose with itself like you are suggesting.

However, I want to keep this PR focused to the EffectHandler itself, and address that problem at a later stage.

MobiusExtras/Test/ConnectableClassTests.swift
jeppes: I see. I only removed these to comply with the warnings-as-errors check. I will re-add these tests and disable the check instead.

MobiusCore/Source/EffectRouterBuilder.swift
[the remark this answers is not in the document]
JensAyton: 👍

MobiusCore/Source/EffectHandlers/EffectHandler.swift
jeppes: I want to avoid the protocol-first approach. Mainly due to how generics are handled. By using a protocol as our base abstraction we'd need to introduce a type-erased `AnyEffectHandler` variant which adds confusion and pollutes the APIs.

I would use this `EffectHandler<Effect, Event>` type as our base abstraction. That way we can write code which depends on the generic idea of an `EffectHandler<F,E>` instead of concrete implementations or type-erasing wrappers.

Using this as our base abstraction also allows users to build their own protocol(s) if they wish, as long as they can supply a function `MyEffectHandlerProtocol -> EffectHandler<..>`. 

This closure based approach still allows you to use classes or structs to hold onto the effect handler state. Your example could, for example, be written as:
```
func myFeatureEffectHandler(thingManager: ThingManager) -> CompositeEffectHandler<MyLoopTypes> {
    return CompositeEffectHandler()
        .addHandler(MyFirstEffectHandler.make())
        .addHandler(MySecondEffectHandler.make(thingManager: thingManager))
}
```

---

I also think there is a deeper issue here. An effect handler should really just be a way of tying a side-effecting construct into a Mobius loop:

For instance, say you have something which can download a song. This "thing" will take some input (which song to download), and produce some output (potentially the downloaded song, download progress updates, errors etc). All of these things exist outside Mobius, and this thing should therefore live outside of Mobius too.

This "thing" is not an effect handler in an of itself. However, it is likely something we want to provide to the loop. An EffectHandler is just an adapter layer which maps such a generic input-output "thing" into a loop. I think the API I am suggesting makes that clearer.

MobiusCore/Source/EffectHandlers/EffectHandler.swift
```diff
@@ -0,0 +1,142 @@
...
+            case .handle: return true
+            default: return false
+            }
+        }
+        self.handleEffect = { effect, dispatch in
+            switch canAccept(effect) {
+            case .handle(let subType):
+                handleEffect(subType, dispatch)
+            default:
+                fatalError("ActionEffectHandler's canAccept is implemented incorrectly. This should not be possible")
```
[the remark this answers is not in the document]
JensAyton: The situation is more subtle than that: this case can only be reached if `canAccept` returns different answers for the same effect on two subsequent calls.

Unfortunately, the current message is a bit optimistic, since you can’t require `canAccept` to be pure.

MobiusCore/Source/EffectHandlers/EffectHandler.swift
```diff
@@ -0,0 +1,142 @@
...
+        disposeFn = disposable
+        canAcceptEffect = { effect in
+            switch canAccept(effect) {
+            case .handle: return true
+            default: return false
+            }
+        }
+        self.handleEffect = { effect, dispatch in
+            switch canAccept(effect) {
+            case .handle(let subType):
```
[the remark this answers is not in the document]
jeppes: I like it! Thanks

MobiusCore/Source/EffectHandlers/EffectHandler.swift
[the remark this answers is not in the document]
jeppes: This has been removed in the latest version.

MobiusCore/Source/EffectHandlers/EffectHandler.swift
```diff
@@ -0,0 +1,146 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+import Foundation
+
+public enum HandleEffect<Type> {
+    case handle(Type)
+    case ignore
+}
```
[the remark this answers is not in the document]
jeppes: Sure it's isomorphic, but I don't think an optional would convey any sort of clarity about what should be done in this method.

MobiusCore/Source/EffectHandlers/EffectHandler.swift
JensAyton: Hmm. I’m seeing quite complex effect handlers in our codebase, but maybe that’s an antipattern. If we actively promote a model where one effect handler binds one effect to one collaborator, the closure-based approach might actually help.

I don’t particularly see the `AnyEffectHandler` issue as a strong argument. This is a well-known wart that Swift developers should get used to, and APIs can be designed so that client code never needs to use `AnyEffectHandler` directly.

In any case, this is more syntax than semantics and we could easily transition from one approach to the other without breaking compatibility, so I guess I’m OK with either for now.

MobiusCore/Source/EffectHandlers/EffectHandler.swift
```diff
@@ -0,0 +1,140 @@
...
+    /// Create a handler for effects which satisfy the `canAccept` parameter function.
+    ///
+    /// - Parameter canHandle: A function which indicates whether to handle an effect. If it returns `.handle(effect)` for a given `effect`, this
+    /// effect handler will handle said effect.
+    /// - Parameter handleEffect: Handle effects which satisfy `canAccept`.
+    /// - Parameter onDispose: Tear down any resources being used by this effect handler.
+    public init<AssociatedValueType>(
+        canHandle: @escaping (Effect) -> HandleEffect<AssociatedValueType>,
+        handleEffect: @escaping (AssociatedValueType, @escaping Consumer<Event>) -> Void,
+        onDispose disposable: @escaping () -> Void
```
JensAyton: On an unrelated note, we need to be very clear with documenting this `onDispose`, since it binds to the lifetime of a `connection` rather than the `connectable` (composite event handler), and may be called multiple times per event handler instance. (This distinction is why I called it `stop` instead of `dispose`)

JensAyton: I think I prefer `can` over `should`, but otherwise I like the second form.

jeppes: Ok, I like it too. I'll change it :)

MobiusCore/Source/EffectHandlers/EffectHandler.swift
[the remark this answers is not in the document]
jeppes: I will remove this type and replace it with an optional given the controversy.

jeppes:
@JensAyton sorry, there was a piece missing from the puzzle. See the latest commit

MobiusCore/Source/BackwardsCompatibility.swift
```diff
@@ -109,3 +109,29 @@ public extension EffectRouterBuilder {
...
+    /// The builder is immutable. When setting various properties, a new instance of a builder will be returned.
+    /// It is therefore recommended to chain the loop configuration functions
+    ///
+    /// Once done configuring the loop you can start the loop using `start(from:)`.
+    ///
+    /// - Parameters:
+    ///   - update: the `Update` function of the loop
+    ///   - effectHandler: an instance conforming to the `ConnectableProtocol`. Will be used to handle effects by the loop
+    /// - Returns: a `Builder` instance that you can further configure before starting the loop
+    @available(*, deprecated, message: "Construct the loop with an `EffectHandler<Effect, Event>` instead")
```
JensAyton: I’m confused. The way to combine `EffectHandler`s is currently still `EffectRouterBuilder`, which produces a `Connectable` rather than an `EffectHandler`.

jeppes: You are completely right. This should not be deprecated.

PR #50, opened 2019-10-29
[request for comment] Remove LoopTypes protocol

mbrandonw:
This PR demonstrates what it would look like to remove the `LoopTypes` protocol. 

The protocol currently serves the purpose of unifying the 3 generics `Model`, `Event` and `Effect` into a single generic. That cleans up some call sites, by doing `Update<T>` instead of `Update<Model, Event, Effect>`, but it also has some drawbacks:

* The protocol indirection can hurt type inference because Swift can't know to look inside `T` to figure out what `Model`, `Event` and `Effect` are. This means you have to specify the generic in places that you typically would not have to:
  ```swift
  // With loop types
  let loggingInitiator = LoggingInitiator<Types>(initiator, logger)
  let loggingUpdate = LoggingUpdate<Types>(update, logger)

  // Without loop types
  let loggingInitiator = LoggingInitiator(initiator, logger)
  let loggingUpdate = LoggingUpdate(update, logger)
  ```
  
  The second commit (b465ea5) of this PR shows a few spots where type inference can be leveraged to simplify the code. Most (maybe all) of these type inferences are impossible with the `LoopTypes`.

* There is currently inconsistent usage of `LoopTypes`, so getting rid of `LoopTypes` would help with that inconsistency. For example, `Next` doesn't use it, probably because it is only generic over `Model` and `Effect`, but not `Event`. However, the `Initiator` type _is_ generic over `LoopTypes` even though it does not need `Effect`. That means `Initiator` technically has an additional, unused generic (also known as a phantom type) that could cause type inference problems if the `Event` cannot be inferred.

* The `LoopTypes` conformance requires creating another type for each Mobius loop, and may be seen as boilerplate. It also causes some indirection if you want to immediately see what types a loop is dealing with, you have to jump to the `LoopTypes` instead of just looking at the type of the loop.

⚠️ It is worth mentioning that this change is not backwards compatible, and would require all existing Mobius features to update. That could be some significant churn.

### DIFF
--- a/MobiusCore/Source/EventProcessor.swift
+++ b/MobiusCore/Source/EventProcessor.swift
@@ -21,14 +21,14 @@ import Foundation
 
 /// Internal class that manages the atomic state updates and notifications of model changes when processing of events via
 /// the Update function.
-class EventProcessor<Types: LoopTypes>: Disposable, CustomDebugStringConvertible {
-    let update: Update<Types>
-    let publisher: ConnectablePublisher<Next<Types.Model, Types.Effect>>
+class EventProcessor<Model, Event, Effect>: Disposable, CustomDebugStringConvertible {
+    let update: Update<Model, Event, Effect>
+    let publisher: ConnectablePublisher<Next<Model, Effect>>
 
     private let queue: DispatchQueue
 
-    private var currentModel: Types.Model?
-    private var queuedEvents = [Types.Event]()
+    private var currentModel: Model?
+    private var queuedEvents = [Event]()
 
     public var debugDescription: String {
         let modelDescription: String
@@ -41,8 +41,8 @@ class EventProcessor<Types: LoopTypes>: Disposable, CustomDebugStringConvertible
     }
 
     init(
-        update: @escaping Update<Types>,
-        publisher: ConnectablePublisher<Next<Types.Model, Types.Effect>>,
+        update: @escaping Update<Model, Event, Effect>,
+        publisher: ConnectablePublisher<Next<Model, Effect>>,
         queue: DispatchQueue
     ) {
         self.update = update
@@ -50,7 +50,7 @@ class EventProcessor<Types: LoopTypes>: Disposable, CustomDebugStringConvertible
         self.queue = queue
     }
 
-    func start(from first: First<Types.Model, Types.Effect>) {
+    func start(from first: First<Model, Effect>) {
         queue.sync(flags: .barrier) {
             currentModel = first.model
 
@@ -64,7 +64,7 @@ class EventProcessor<Types: LoopTypes>: Disposable, CustomDebugStringConvertible
         }
     }
 
-    func accept(_ event: Types.Event) {
+    func accept(_ event: Event) {
         queue.async(flags: .barrier) {
             if let current = self.currentModel {
                 let next = self.update(current, event)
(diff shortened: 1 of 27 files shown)


MobiusCore/Source/EventProcessor.swift
```diff
@@ -21,14 +21,14 @@ import Foundation
 
 /// Internal class that manages the atomic state updates and notifications of model changes when processing of events via
 /// the Update function.
-class EventProcessor<Types: LoopTypes>: Disposable, CustomDebugStringConvertible {
-    let update: Update<Types>
-    let publisher: ConnectablePublisher<Next<Types.Model, Types.Effect>>
+class EventProcessor<Model, Event, Effect>: Disposable, CustomDebugStringConvertible {
+    let update: Update<Model, Event, Effect>
```
mbrandonw: Here we can see that instead of doing `Types: LoopTypes` we have to do `Model, Event, Effect`, but perhaps that isn't too different. I believe it is more typical for library level code to have to worry about the generics, whereas client code can usually use type inference to not have to worry about it.

Also worth noting that we get to remove some `Types.` noise when needing to refer to the generics inside `Types`.

MobiusCore/Source/Mobius.swift
```diff
@@ -45,36 +39,36 @@ public extension Mobius {
...
+        return Builder(
             update: update,
             effectHandler: effectHandler,
             initiator: { First(model: $0) },
-            eventSource: AnyEventSource<T.Event>({ _ in AnonymousDisposable(disposer: {}) }),
+            eventSource: AnyEventSource({ _ in AnonymousDisposable(disposer: {}) }),
             eventQueue: DispatchQueue(label: "event processor"),
             effectQueue: DispatchQueue(label: "effect processor", attributes: .concurrent),
-            logger: AnyMobiusLogger(NoopLogger<T>())
+            logger: AnyMobiusLogger(NoopLogger())
```
mbrandonw: Here is an example of type inference that wasn't possible before, as well as the `eventSource` line above.

MobiusCore/Source/MobiusLogger.swift
```diff
@@ -84,11 +85,7 @@ class NoopLogger<Types: LoopTypes>: MobiusLogger {
 }
 
 /// Type-erased `MobiusLogger`.
-public class AnyMobiusLogger<Types: LoopTypes>: MobiusLogger {
-    public typealias Model = Types.Model
-    public typealias Event = Types.Event
-    public typealias Effect = Types.Effect
```
mbrandonw: We can also now leverage type inference for picking up these associated types.

MobiusCore/Source/Mobius.swift
```diff
@@ -19,15 +19,9 @@
...
-    associatedtype Model
-    associatedtype Event
-    associatedtype Effect
-}
-
-public typealias Update<T: LoopTypes> = (T.Model, T.Event) -> Next<T.Model, T.Effect>
+public typealias Update<Model, Event, Effect> = (Model, Event) -> Next<Model, Effect>
 
-public typealias Initiator<T: LoopTypes> = (T.Model) -> First<T.Model, T.Effect>
+public typealias Initiator<Model, Effect> = (Model) -> First<Model, Effect>
```
mbrandonw: Turns out the initiator isn't using `Event` even though it is generic over it. Such phantom types can hurt type inference.

PR #51, opened 2019-10-31
Proposal: Remove Mobius Playground

jeppes:
I propose that we remove the Mobius Playground.
It doesn't work and playgrounds are too unstable to be useful anyway.

I would propose that if we want something similar to this, we should write it as a series of unit tests instead.


JensAyton:
But I want it to work. ☹️

Also merge conflicts.

jeppes:
@JensAyton I think we should integrate it with our CI if we want to keep it

PR #52, opened 2019-10-31
Composable effect handlers

jeppes:
Add the ability to compose `EffectHandler`s.
The semantics (when connecting to a loop) are the same as with the `EffectRouterBuilder`. I.e., exactly one `EffectHandler` needs to handle each effect.

This also deprecates the `EffectRouterBuilder`.


MobiusCore/Source/EffectHandlers/EffectHandlerComposition.swift
```diff
@@ -0,0 +1,62 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+public extension EffectHandler {
+    /// Merge a group of `EffectHandler`s.
+    /// Note: Any given `Effect` can be handled by at most  __1__ `EffectHandler`.
```
pettermahlen: This could be phrased more strongly, indicating that  a runtime error will happen if there are collisions.

MobiusCore/Source/EffectHandlers/EffectHandler.swift
```diff
@@ -41,21 +38,31 @@ final public class EffectHandler<Effect, Event> {
...
+               } else {
+                   return nil
+               }
+            },
+            stopHandling: disposable
+        )
+    }
+
+    init(
+        handleEffect: @escaping (Effect) -> ((@escaping Consumer<Event>) -> Void)?,
```
pettermahlen: I feel like `handleEffect` is a confusing name, and it's making the static merging `handleEffect` function harder to understand. AFAIU, this function isn't really handling the effect, but determining if it can be handled by this effect handler, and if so, what code should run in order to handle it. The actual execution of that code happens later. If that's not correct (and side effects can happen during the evaluation of `handleEffect`), then the merging function looks broken. 

Maybe something like `selectHandler` could be a better name? It's not completely clear to me that this class is really an `EffectHandler` either, it may be more of an effect router. Either things are a little muddled here, or there are things I'm not understanding.

jeppes:

MobiusCore/Source/Mobius.swift
```diff
@@ -218,3 +195,26 @@ class LoggingUpdate<Model, Event, Effect> {
...
+    /// We must therefore guarantee this property at the point where we connect our `EffectHandler` to the loop.
+    var asLoopConnectable: AnyConnectable<Effect, Event> {
+        return AnyConnectable { dispatch -> Connection<Effect> in
+            let effectHandlerConnection = self.connect(dispatch)
+            return Connection(
+                acceptClosure: { effect in
+                    if let executeEffect = self.canHandle(effect) {
+                        executeEffect(dispatch)
+                    } else {
+                        MobiusHooks.onError("More than one effect handler handling effect: \(effect)")
```
pettermahlen: Isn't this also possibly 'no effect handler handling effect'?

jeppes: Yes, and that is actually the only error it can be. `More than one effect handler` can only be caught in the composition of EffectHandlers, not when attaching to the loop.

jeppes: Fixed in the relevant commit.

MobiusCore/Source/EffectHandlers/EffectHandlerComposition.swift
jeppes: Good point

jeppes:
I will close this pending some work on simplifying the current structure of EffectHandler. It is tied to the way `Connectable`s used to work, and I suspect that this can be simplified.

PR #59, opened 2019-11-05
Update tooling and scripts

dflems:
> edit: GitHub is sad today: https://www.githubstatus.com - I'll wait a bit and retrigger

This is falling over in GH, likely due to https://github.com/Quick/Nimble/issues/702


jeppes:
thanks @dflems

PR #61, opened 2019-11-06
Duplicate `Lock` in MobiusExtras to fix SPM Build

BalestraPatrick:
SPM doesn't like the fact that we reference `Lock` from [`ConnectableClass`](https://github.com/spotify/Mobius.swift/blob/3d8832f3ac11d9ca66193b7149ad634be94f69e9/MobiusExtras/Source/ConnectableClass.swift#L40) even though it's private. SPM fails to build Mobius when used as a SPM package, but this PR fixes it.

I'm not even sure how this compiles fine inside the project right now, but with @jeppes we agreed that the best short-term solution is to duplicate the code to avoid exposing `Lock` as a public type.


dflems:
Fixed in #64

JensAyton:
The way it worked was to have two references to the same file.

PR #68, opened 2019-11-12
Deprecate direct construction of MobiusController

JensAyton:
Obvious next step is obvious. @jeppes

### DIFF
--- a/MobiusCore/Source/Mobius.swift
+++ b/MobiusCore/Source/Mobius.swift
@@ -154,13 +154,13 @@ public extension Mobius {
         /// Create a `MobiusController` from the builder
         ///
         /// - Parameters:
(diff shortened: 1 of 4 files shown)


MobiusCore/Source/Mobius.swift
```diff
@@ -154,13 +154,13 @@ public extension Mobius {
         /// Create a `MobiusController` from the builder
         ///
         /// - Parameters:
-        ///   - defaultModel: The initial default model of the `MobiusController`
+        ///   - initialModel: The initial default model of the `MobiusController`
```
rastersize: Should we remove the “default” here in the comment as well?

JensAyton: It is the “initial default model” of the controller, since a new default model can be set with `replaceModel`.

PR #72, opened 2019-11-13
EventSource as Connectable<Model, Event>

jeppes:
It should be possible to create an `EventSource` as a `Connectable<Model, Event>`.
This will make it easier to compose loops (since event sources can be triggered on a per sub-domain basis), and it will align the Swift version with the Java version.

There are some things we need to discuss:
1. Should `EventSource<Event>` be removed entirely in favor of `EventSource<Model, Event>`?
2. Should we create an approachable way of creating thread-safe `EventSource`s? This PR uses the EffectRouter to do so. This is doable, but definitely not obvious.

### DIFF
--- a/Mobius.xcodeproj/project.pbxproj
+++ b/Mobius.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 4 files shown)


JensAyton:
Hmm, apparently I can’t retarget this PR. :-/

jeppes:
I'll move it @JensAyton

PR #73, opened 2019-11-13
Event source as connectable

jeppes:
It should be possible to create an EventSource as a Connectable<Model, Event>.
This will make it easier to compose loops (since event sources can be triggered on a per sub-domain basis), and it will align the Swift version with the Java version.

There are some things we need to discuss:

Should EventSource<Event> be removed entirely in favor of EventSource<Model, Event>?
Should we create an approachable way of creating thread-safe EventSources? This PR uses the EffectRouter to do so. This is doable, but definitely not obvious.

### DIFF
--- a/Mobius.xcodeproj/project.pbxproj
+++ b/Mobius.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		0280FB55237C1476007489A2 /* EventSourceConnectableTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0280FB53237C1450007489A2 /* EventSourceConnectableTests.swift */; };
 		02BED1BB21DD20D20093FB47 /* ConnectableContramap.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5B9CE80421197FE000DB79A7 /* ConnectableContramap.swift */; };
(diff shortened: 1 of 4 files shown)


JensAyton:
I remain skeptical to the value of `Connectable` as an exposed thing. Also, if an event source can receive a model, how is it different from an observer/view?

jeppes:
@JensAyton 
> I remain skeptical to the value of Connectable as an exposed thing

I agree. I'd like to create something similar to `EffectHandler` but for event sources.

> how is it different from an observer/view?

It's not. Or, at least, I don't think it should be. The UI is just another source of events after all.

pettermahlen:
I guess the `EffectRouter` could be generalised to be a `Router`, or the same ideas used. Then, instead of having multiple observers of the model, you could set up routing of models to different observers. It would be a different kind of routing, because we'd definitely like to have the same model sent to multiple (all!) observers. Also, model observers tend to want to have the 'current model' immediately on subscribe, so there are some differences between them. So there are similarities and differences between the event-effect-event loop part and the event-model-event part. We did discuss making model updates into a kind of effect, though, which would combine the two. A reason to avoid that is that the `UpdateModel` effect would have to be included in all effect types, which doesn't play well with using sum types for them. We'd need some kind of union of 'mobius-internal' and 'domain-specific' types for that. Might be a case for update function composition/nesting.

I think I'd personally like to park this for a bit and see where we end up with the other changes we're making. I think we'll learn things that might give us new ideas around this.

jeppes:
@pettermahlen I'm ok with parking this for now, but I would like to revisit this before 1.0

Generalizing the router seems interesting. There is really nothing Effect specific there except the names. I think we should explore this idea further.

Anyway, I will close this for now.

PR #75, opened 2019-11-13
Wippedy-doo-dah

JensAyton:
Continuation of #45.

MobiusController possibly done? Event sources and events from views are scheduled on the loop queue as intended. Need test cases for events from synchronous and asynchronous effects.

Also, events from views are redundantly sent to the loop queue in both `AsyncDispatchQueueConnectable` and `eventFilter`. `AsyncDispatchQueueConnectable` is probably unnecessary.

`dispatchPrecondition` is awesome, maybe there are some more places I can use that.


JensAyton:
This seems to work as intended now, but will be easier to review if rebased after #78, #79 and #81 are merged.

PR #76, opened 2019-11-14
Remove ConnectableClass deprecation

jeppes:
ConnectableClass still has use cases outside of handling effects. We should remove the deprecation until we have a true replacement


jeppes:
@JensAyton

PR #77, opened 2019-11-18
Revert "Inline Lock functionality in ConnectableClass"

dflems:
This reverts commit 9823762c2e736967b6856947bbc8786551aac374.

@BalestraPatrick - Just duplicating the class worked fine for me locally. Why was this necessary?


BalestraPatrick:
That was my fix initially, but @jeppes had a comment about that: https://github.com/spotify/Mobius.swift/pull/61#pullrequestreview-312308734

dflems:
ok!

PR #78, opened 2019-11-22
Ergonomics tweaks

JensAyton:
Conveniences for loggers:
* Loggers aren’t required to implement methods they don’t care about
* `ConsoleLogger` generalized to `SimpleLogger` which takes a print function (we have this in our own codebase already)

Also, the debug description for MobiusLoop now includes type info even when the loop has been disposed.

@jeppes


MobiusExtras/Source/SimpleLogger.swift
```diff
@@ -20,36 +20,41 @@
 import Foundation
 import MobiusCore
 
-public class ConsoleLogger<Model, Event, Effect>: MobiusLogger {
+public class SimpleLogger<Model, Event, Effect>: MobiusLogger {
     private let prefix: String
+    private let consumer: Consumer<String>
 
-    public init(tag: String = "Mobius") {
+    public init(tag: String = "Mobius", consumer: @escaping Consumer<String> = { print($0) }) {
```
JensAyton: On further consideration, it would be better to take a `TextOutputStream` and use `print(_:to:)`

JensAyton: Turns out not to be a great idea

JensAyton:
Merged in master, and also merged in #84 (already approved) to reduce the amount of merge/rebase dancing needed

PR #79, opened 2019-11-22
Update deployment targets

JensAyton:
On my working branch, I’m seeing CI failures for using iOS 10/macOS 10.12 functionality. I don’t think we have any intention of supporting pre-2016 OSes.

@dflems @jeppes


.gitignore
```diff
@@ -4,6 +4,7 @@
 ## Build generated
 build/
 DerivedData/
+Mobius.xcodeproj/xcshareddata/xcschemes/ZZZ_MOBIUS_ALL.xcscheme
```
dflems: ah should have done this from the start :)

PR #81, opened 2019-11-22
Cleanup for threading changes

JensAyton:
Separate out various orthogonal concerns from #75 to make it easier to review later.

@jeppes @togi @pettermahlen

### DIFF
--- a/MobiusCore/Source/MobiusController.swift
+++ b/MobiusCore/Source/MobiusController.swift
@@ -145,7 +145,7 @@ public class MobiusController<Model, Event, Effect> {
     /// if it's not running.
     ///
     /// - Returns: a model with the state of the controller
-    public func getModel() -> Model {
+    public var model: Model {
         return lock.synchronized {
             loop?.latestModel ?? modelToStartFrom
         }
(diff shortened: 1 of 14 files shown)


JensAyton:
@dflems Any opinion on the linter changes? It’s our internal one with the localization regex rule removed and different exception paths.

MobiusCore/Source/MobiusController.swift
```diff
@@ -33,7 +33,7 @@ public class MobiusController<Model, Event, Effect> {
     private var modelToStartFrom: Model
 
     /// A Boolean indicating whether the MobiusLoop is running or not.
-    public var isRunning: Bool {
+    public var running: Bool {
```
pettermahlen: These two changes are made in a backwards-incompatible way instead of deprecating the old version and adding a new, as has been done before. I'm fine with that, I think, but I wanted to note that that's a difference in approach. I'm fine with it because I think we're going to make a number of backwards-incompatible changes for 1.0.

JensAyton: Look again. :-)

pettermahlen: Clever trick to fool your reviewers with! :)

PR #82, opened 2019-11-25
Update nimble to resolve test issues

marmelroy:
v8.0.0 of Nimble doesn't build with Xcode 11 with the Swift Package Manager -  we need to bump to 8.0.4.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -16,7 +16,7 @@ let package = Package(
         .library(name: "MobiusTest", targets: ["MobiusTest"]),
     ],
     dependencies: [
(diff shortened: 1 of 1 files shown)


marmelroy:
For more context on this:
- cwlpreconditiontesting has a swift package manager bug fix in v2.0.0.
- nimble's 8.0.4 tagged release is pointing at cwlpreconditiontesting's old version.
- Nimble's master branch is pointing at the right version of cwlpreconditiontesting.
- I have created a temporary workaround on my branch but I don't think we should merge it here. 
- I've opened an issue on Nimble asking for a tagged release that will automatically fix it.

PR #83, opened 2019-11-25
Enable TSAN

JensAyton:
It turns out that our effect handler tests had a bunch of data races. Let’s not have that.

@pettermahlen @dflems

### DIFF
--- a/Tools/ci.sh
+++ b/Tools/ci.sh
@@ -23,7 +23,7 @@ has_command xcpretty || fail "xcpretty must be installed"
 
 # until then using SPM to unblock CI:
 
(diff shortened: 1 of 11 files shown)


Tools/ci.sh
```diff
@@ -23,7 +23,7 @@ has_command xcpretty || fail "xcpretty must be installed"
 
 # until then using SPM to unblock CI:
 
-swift test --enable-code-coverage || fail "SPM Test Failed"
+swift test --enable-code-coverage --sanitize=thread || fail "SPM Test Failed"
```
dflems: :+1:

PR #84, opened 2019-11-27
Logging adaptor cleanup

JensAyton:
Incidental cleanup that happened while I was investigating threading bugs

@togi


JensAyton:
Merged into #78

PR #89, opened 2019-12-06
Loop controller state machine

JensAyton:
Refactor the loop controller as an explicit state machine, separating the “business logic” of starting and stopping loops from the representation of state transitions and the queue management.

This fixes two known problems in #85:
* The `running` state in `flipEventsToLoopQueue` is explicitly synchronized. (This is basically the old `AtomicBool`, but generic. Oh well. I’m sure we’ll get real atomics one of these years.)
* Transitional states are explicitly modelled, which addresses the problem where issuing an event immediately on subscribing to an event source would crash the MobiusController. The disabled regression test for this is enabled (it was the only disabled test).

@pettermahlen @togi


MobiusCore/Source/MobiusController.swift
```diff
@@ -157,22 +136,17 @@ public final class MobiusController<Model, Event, Effect> {
...
+        state.syncTransitionToStopped(error: "cannot stop a controller that isn't running") { runningState in
+            let model = runningState.loop.latestModel
 
-            loop.dispose()
-            viewConnection?.dispose()
+            runningState.loop.dispose()
+            runningState.viewConnection.dispose()
 
-            self.loop = nil
+            return StoppedState(modelToStartForm: model, viewConnectable: runningState.viewConnectable)
```
pettermahlen:
```suggestion
            return StoppedState(modelToStartFrom: model, viewConnectable: runningState.viewConnectable)
```

MobiusCore/Source/MobiusController.swift
```diff
@@ -183,13 +157,8 @@ public final class MobiusController<Model, Event, Effect> {
...
     public func replaceModel(_ model: Model) {
-        synchronized {
-            guard loop == nil else {
-                MobiusHooks.onError("cannot replace the model of a running loop")
-                return
-            }
-
-            modelToStartFrom = model
+        state.syncMutateStopped(error: "cannot replace the model of a running loop") { state in
+            state.modelToStartForm = model
```
pettermahlen:
```suggestion
            state.modelToStartFrom = model
```

MobiusCore/Source/MobiusController.swift
```diff
@@ -199,13 +168,152 @@ public final class MobiusController<Model, Event, Effect> {
...
+        private var stoppedState: StoppedState?
+        private var runningState: RunningState?
+
+        enum Snapshot {
+            case stopped(StoppedState)
+            case running(RunningState)
+        }
+
+        init(model: Model, queue: DispatchQueue) {
+            stoppedState = StoppedState(modelToStartForm: model, viewConnectable: nil)
```
pettermahlen:
```suggestion
            stoppedState = StoppedState(modelToStartFrom: model, viewConnectable: nil)
```

MobiusCore/Source/MobiusController.swift
```diff
@@ -48,21 +46,20 @@ public final class MobiusController<Model, Event, Effect> {
         viewQueue: DispatchQueue
     ) {
         let actualLoopQueue = DispatchQueue(label: "MobiusController \(Model.self)", target: loopQueue)
-        modelToStartFrom = initialModel
-        self.loopQueue = actualLoopQueue
         self.viewQueue = viewQueue
 
-        var running: () -> Bool = { false }
+        let state = State(model: initialModel, queue: actualLoopQueue)
```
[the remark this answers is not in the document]
JensAyton: Capturing `self` in `flipEventsToLoopQueue` is impossible because of definite initialization requirements, but I’ll add a comment to clarify the overall structure here

MobiusCore/Source/MobiusController.swift
```diff
@@ -199,13 +168,152 @@ public final class MobiusController<Model, Event, Effect> {
     ///
```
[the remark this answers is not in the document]
JensAyton: It is, this still wraps a `loopQueue.sync`

MobiusCore/Source/MobiusController.swift
```diff
@@ -199,13 +168,152 @@ public final class MobiusController<Model, Event, Effect> {
...
+            switch rawState.value {
+            case .stopped:
+                return false
+            case .transitioningToRunning, .running, .transitioningToStopped:
+                return true
+            }
+        }
+
+        func syncRead<T>(_ closure: (Snapshot) throws -> T) rethrows -> T {
+            return try loopQueue.sync {
```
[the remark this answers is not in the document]
JensAyton: The distinction only exists in the `MobiusController` initializer; I’ll rename and comment for clarity

MobiusCore/Source/MobiusController.swift
```diff
@@ -199,13 +168,152 @@ public final class MobiusController<Model, Event, Effect> {
...
+            loopQueue.sync {
+                switch snapshot() {
+                case .running:
+                    MobiusHooks.onError(error)
+                case .stopped(let stoppedState):
+                    rawState.value = .transitioningToRunning
+                    if let runningState = transition(stoppedState) {
+                        become(running: runningState)
+                    } else {
+                        rawState.value = .stopped
```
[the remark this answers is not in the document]
JensAyton: I’ll add a doc comment. This happens if the transition function returns nil, which happens after an `onError` (since the error hook is non-terminating).

JensAyton:
I think I’ve addressed all the comments, and also fixed a threading logic bug (events sent from view connections would call `MobiusLoop.dispatchEvent` on the view queue, which is invalid and trips the concurrent access detector, even though it would later be flipped over to the loop queue before actually executing.)

MobiusCore/Source/MobiusController.swift
```diff
@@ -33,36 +33,67 @@ public final class MobiusController<Model, Event, Effect> {
...
     init(
         builder: Mobius.Builder<Model, Event, Effect>,
         initialModel: Model,
-        loopQueue: DispatchQueue,
+        loopQueue loopTargetQueue: DispatchQueue,
         viewQueue: DispatchQueue
     ) {
-        let actualLoopQueue = DispatchQueue(label: "MobiusController \(Model.self)", target: loopQueue)
+        // The internal loopQueue is a serial queue targeting the provided queue, so that targeting a concurrent queue
+        // doesn’t result in concurrent work on the underlying MobiusLoop.
```
pettermahlen: Is 'targeting' a synchronous operation? What I mean is, is there a risk of the following:

1. Event A arrives and is placed on `loopQueue`
1. Event B arrives and is placed on `loopQueue`
1. Event A starts processing. 
1. Event B starts processing concurrently with A.

MobiusCore/Source/MobiusController.swift
```diff
@@ -33,36 +33,67 @@ public final class MobiusController<Model, Event, Effect> {
...
+        let state = State(model: initialModel, queue: loopQueue)
         self.state = state
-
+        /*
+         Note: `flipEventsToLoopQueue` should not and in fact cannot capture `self`, because it’s used to initialize
+         `self.loopFactory`. It can however capture `state`.
+
+         In fact, here’s the ownership graph after initing:
+                     ┏━━━━━━━━━━━━┓
+                ┌────┨ controller ┠────────┐
```
pettermahlen: Nice graph! The ASCII graphics are a little messed up in the GHE browser; I assume they look fine in XCode?

MobiusCore/Source/MobiusController.swift
```diff
@@ -234,13 +268,16 @@ public final class MobiusController<Model, Event, Effect> {
...
         func syncRead<T>(_ closure: (Snapshot) throws -> T) rethrows -> T {
+            dispatchPrecondition(condition: .notOnQueue(loopQueue))
             return try loopQueue.sync {
                 try closure(snapshot())
             }
         }
 
-        func syncMutateStopped(error: String, _ closure: (inout StoppedState) -> Void) {
+        /// Mutate the stopped state, assuming we’re currently stopped. If not, fail with the provided error message.
+        func mutateIfStopped(error: String, _ closure: (inout StoppedState) -> Void) {
```
pettermahlen: Would it be a good idea to rename `error` into `orError`? That would clarify the meaning of the parameter to me, but I'm not sure if it is unidiomatic. Or if 'unidiomatic' is a word for that matter. :)

MobiusCore/Source/ConnectableConvenienceClasses/AsyncDispatchQueueConnectable.swift
```diff
@@ -19,27 +19,40 @@
 
 import Foundation
 
+/// A connectable adapter which imposes asynchronous dispatch blocks around calls in both directions.
+///
+/// * Consumers passed to `connect` will be executed on the provided `consumerQueue`
+/// * The underlying connectable’s connections’ `acceptClosure` and `disposeClosure` will be executed on the provided
```
pettermahlen: It looks to me like the `acceptClosure` is executed on the `acceptQueue`?

pettermahlen: and the dispose, too, so this is probably a typo wrt the queue name?

MobiusCore/Source/MobiusController.swift
```diff
@@ -44,35 +45,40 @@ public final class MobiusController<Model, Event, Effect> {
...
                            ┃ loopQueue ┃
                            ┗━━━━━━━━━━━┛
+
+         In order to construct this bottom-up and fulfil definitive initialization requirements, state and loopQueue are
+         duplicated in local variables.
          */
 
+        // The internal loopQueue is a serial queue targeting the provided queue, so that targeting a concurrent queue
+        // doesn’t result in concurrent work on the underlying MobiusLoop. This behaviour is documented on
+        // `Mobius.Buildre.makeController`.
```
pettermahlen: `Buildre`

MobiusCore/Source/MobiusController.swift
JensAyton: Looks fine in my Github :-/

The vertical spacing is a bit nicer in Xcode, unless of course you use a proportional font for comments. My other option was to embed some Graphviz code…

MobiusCore/Source/MobiusController.swift
JensAyton: When a serial dispatch queue `Q` (which is the default type) feeds into a concurrent dispatch queue `P`, blocks submitted to `Q` are guaranteed to be executed serially with respect to each other, although they may be concurrent with other blocks submitted to `P`. So step 4 in your scenario can’t happen.

MobiusCore/Source/MobiusController.swift
JensAyton: Either way works for me

MobiusCore/Source/MobiusController.swift
pettermahlen: Great!

PR #90, opened 2019-12-06
Deprecate raw loop initiator

JensAyton:
For consistency with Android. We can remove the backwards compatibility and simplify the implementation after 0.3.

@togi

### DIFF
--- a/MobiusCore/Source/Mobius.swift
+++ b/MobiusCore/Source/Mobius.swift
@@ -21,7 +21,7 @@ import Foundation
 
 public typealias Update<Model, Event, Effect> = (Model, Event) -> Next<Model, Effect>
 
-public typealias Initiator<Model, Effect> = (Model) -> First<Model, Effect>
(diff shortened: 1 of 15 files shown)


MobiusCore/Source/Mobius.swift
```diff
@@ -47,7 +47,7 @@ public extension Mobius {
         return Builder(
             update: update,
             effectHandler: effectHandler,
-            initiator: { First(model: $0) },
+            initiate: nil,
```
jeppes: Not a big deal, but it seems the convention in this file is to stub out things which are not provided rather than making them `nil`.
I think `nil` is the way we should go, but it does look inconsistent within the context of the existing file.

JensAyton: As implemented, initiators and initial effects are mutually exclusive (so I don’t have to define an ordering rule). Since closures don’t have identity, this means we have to track whether an explicit initiator was set, and an optional is the cleanest way.

PR #91, opened 2019-12-06
[WIP] test assertion cleanup

JensAyton:
The use of `MobiusHooks` to filter assertions so we can test them is quite unpleasant. As it turns out, it was also causing tests to pass which shouldn’t have, in some cases because of hooks being set and not cleared in other tests (which made the results different depending on whether you ran all the tests or just one test suite).

Fortunately (?), Nimble now incorporates a [horrible, horrible hack](http://www.cocoawithlove.com/blog/2016/02/02/partial-functions-part-two-catching-precondition-failures.html) which allows for ergonomic testing of assertions. Yay!

The first three commits are #89 and can be ignored here.

### DIFF
--- a/MobiusCore/Source/Lock.swift
+++ b/MobiusCore/Source/Lock.swift
@@ -22,12 +22,36 @@ import Foundation
 struct Lock {
     private let lock = NSRecursiveLock()
 
-    func synchronized<R>(closure: () -> R) -> R {
+    func synchronized<Result>(closure: () throws -> Result) rethrows -> Result {
         lock.lock()
         defer {
             lock.unlock()
         }
 
-        return closure()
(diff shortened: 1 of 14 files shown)


JensAyton:
Now without Objective-C thunks

JensAyton:
Looks like CwlPreconditionTesting bits are exposed as part of Nimble when using Carthage, but not with SPM. Will dig deeper next week.

JensAyton:
```
fatalError("The throwAssertion Nimble matcher does not currently support Swift CLI." +
            " You can silence this error by placing the test case inside an #if !SWIFT_PACKAGE" +
            " conditional statement")
```

Welp.

PR #92, opened 2019-12-12
RFC: Wrap Update in struct

stephencelis:
This PR follows from discussion with @anawara and @jeppes. It adds a struct wrapper for the update function in order to allow for better ergonomics around enhancing update functions with additional functionality. To provide an example of such enhancement, we've taken an internal helper class, `LoggingUpdate`, and shown how the same functionality can be achieved through a helper method on `Update`.

This PR is a breaking change. We could introduce a non-breaking version but would need to introduce a different name than `Update`.


stephencelis:
@JensAyton @pettermahlen Wondering if you have any thoughts!

pettermahlen:
I don't have the Swift chops to have an informed opinion unfortunately.

MobiusCore/Source/MobiusLoop.swift
```diff
@@ -104,7 +104,7 @@ public final class MobiusLoop<Model, Event, Effect>: Disposable, CustomDebugStri
 
     // swiftlint:disable:next function_parameter_count
     static func createLoop<C: Connectable>(
-        update: @escaping Update<Model, Event, Effect>,
+        update: Update<Model, Event, Effect>,
```
jeppes: Can we add the follow overload for this:
```swift
static func createLoop<C: Connectable>(
    update: @escaping (Model, Event) -> Next<Model, Effect>,
    ...
)
```

That way most of the existing code should still work as far as I can tell. The only cases which would break would be the onces which explicitly referenced the `Update` typealias.

JensAyton: Agreed

MobiusCore/Source/Mobius.swift
```diff
@@ -19,7 +19,17 @@
 
 import Foundation
 
-public typealias Update<Model, Event, Effect> = (Model, Event) -> Next<Model, Effect>
+public struct Update<Model, Event, Effect> {
+    private let update: (Model, Event) -> Next<Model, Effect>
+
+    public init(update: @escaping (Model, Event) -> Next<Model, Effect>) {
+        self.update = update
+    }
```
JensAyton: Do we also want a wrapper `init(update: @escaping(inout Model, Event) -> [Effect])`, or is that coming later? :-)

MobiusCore/Source/MobiusLoop.swift
stephencelis: This `createLoop` appears to be internal. Should I add the overload to `Mobius.loop`?

MobiusCore/Source/Mobius.swift
stephencelis: I'd be happy to do a followup PR since more internals will likely change :)

stephencelis:
@jeppes UpdateSpec.swift was updated here: https://github.com/spotify/Mobius.swift/pull/92/files#diff-7f155f3e4eefa81fdf28209e0613b21f

Is there something that's missing?

MobiusCore/Source/MobiusLoop.swift
JensAyton: That makes a lot more sense. :-)

stephencelis: Done!

stephencelis:
:wave: @jeppes Anything holding this up?

stephencelis:
@jeppes :wave: Can you merge please? (I don't have the ability.)

PR #93, opened 2019-12-16
[RFC] Use inout signature for update function

mbrandonw:
(**NB**: This PR is technically based off of #92, but until #92 is merged we won't be able to see the minimal diff here. To see the true diff click [here](https://github.com/stephencelis/Mobius.swift/compare/update-struct...stephencelis:inout))

## What

This changes the signature of `Update` to be of the form `(inout Model, Event) -> [Effect]` instead of `(Model, Event) -> (Model, [Effect])`.

Although it may seem that allowing mutation in the reducer is not something we want, it is in fact no different to returning a new model, thanks to Swift's value types. The mutation is locally scoped to just the invocation of this function, as opposed to reference type mutations, which can travel far and are not controlled.

There are two main benefits to doing this:

### Simpler reducers

Implementations of reducers can be simpler. Currently you must construct a new model to return from a reducer, which can be done in 3 different ways:

* Construct it from scratch each time:
     ```swift
     func update(model: MyModel, event: Event) -> Next<Model, Effect> {
       return Next(MyModel(id: model.id, name: model.name, isAdmin: true))
     }
     ```
    This can get really tedious for large models.
* Adopt a "builder" pattern for models:
     ```swift
     func update(model: MyModel, event: Event) -> Next<Model, Effect> {
       return Next(model.builder().isAdmin(true).build())
     }
     ```
     This requires some kind of code generation to give us access to these builder methods.
* Make a mutable copy, and then return it at the end:
     ```swift
     func update(model: MyModel, event: Event) -> Next<Model, Effect> {
       var model = model
       model.isAdmin = true
       return Next(model)
     }
     ```
    This is boilerplate that will be done in every reducer, and either requires shadowing the model or potentially having two different versions of the model in the scope of this function.

Alternatively, with `inout` we can just apply the mutations directly without any set up, and then return the effects:

```swift
func update(model: MyModel, event: Event) -> [Effect] {
  model.isAdmin = true
  return []
}
```

### More efficient reducers

Reducers no longer need to make entirely new copies of the model every time an event is dispatched. This may not always be a concern, but for very large models it could become a problem.

### DIFF
--- a/MobiusCore/Source/Mobius.swift
+++ b/MobiusCore/Source/Mobius.swift
@@ -21,14 +21,22 @@ import Foundation
 
 /// A wrapper around an update function.
 public struct Update<Model, Event, Effect> {
-    private let update: (Model, Event) -> Next<Model, Effect>
+    private let update: (inout Model, Event) -> [Effect]
 
-    public init(_ update: @escaping (Model, Event) -> Next<Model, Effect>) {
+    public init(_ update: @escaping (inout Model, Event) -> [Effect]) {
         self.update = update
     }
 
-    public func update(model: Model, event: Event) -> Next<Model, Effect> {
-        return self.update(model, event)
+    public static func create(_ update: @escaping (Model, Event) -> Next<Model, Effect>) -> Update {
+        return Update { model, event in
+            let next = update(model, event)
+            model = next.model ?? model
+            return next.effects
+        }
+    }
+
+    public func update(into model: inout Model, event: Event) -> [Effect] {
+        return self.update(&model, event)
     }
 }
 
@@ -76,7 +84,7 @@ public extension Mobius {
         effectHandler: C
     ) -> Builder<Model, Event, Effect> where C.InputType == Effect, C.OutputType == Event {
         return self.loop(
-            update: Update(update),
+            update: Update.create(update),
             effectHandler: effectHandler
         )
     }
--- a/MobiusCore/Source/Mobius.swift
+++ b/MobiusCore/Source/Mobius.swift
@@ -65,6 +65,22 @@ public extension Mobius {
         )
     }
 
+    /// A convenience version of `loop` that takes an unwrapped update function.
+    ///
+    /// - Parameters:
+    ///   - update: the update function of the loop
+    ///   - effectHandler: an instance conforming to the `ConnectableProtocol`. Will be used to handle effects by the loop
+    /// - Returns: a `Builder` instance that you can further configure before starting the loop
+    static func loop<Model, Event, Effect, C: Connectable>(
+        update: @escaping (Model, Event) -> Next<Model, Effect>,
+        effectHandler: C
+    ) -> Builder<Model, Event, Effect> where C.InputType == Effect, C.OutputType == Event {
+        return self.loop(
+            update: Update(update),
+            effectHandler: effectHandler
+        )
+    }
+
     struct Builder<Model, Event, Effect> {
         private let update: Update<Model, Event, Effect>
         private let effectHandler: AnyConnectable<Effect, Event>
--- a/MobiusCore/Source/Mobius.swift
+++ b/MobiusCore/Source/Mobius.swift
@@ -19,7 +19,17 @@
 
 import Foundation
 
-public typealias Update<Model, Event, Effect> = (Model, Event) -> Next<Model, Effect>
+public struct Update<Model, Event, Effect> {
+    private let update: (Model, Event) -> Next<Model, Effect>
+
+    public init(update: @escaping (Model, Event) -> Next<Model, Effect>) {
+        self.update = update
+    }
+
+    public func update(model: Model, event: Event) -> Next<Model, Effect> {
+        return self.update(model, event)
+    }
+}
 
 public typealias Initiator<Model, Effect> = (Model) -> First<Model, Effect>
 
@@ -41,7 +51,7 @@ public extension Mobius {
     ///   - effectHandler: an instance conforming to the `ConnectableProtocol`. Will be used to handle effects by the loop
     /// - Returns: a `Builder` instance that you can further configure before starting the loop
     static func loop<Model, Event, Effect, C: Connectable>(
-        update: @escaping Update<Model, Event, Effect>,
+        update: Update<Model, Event, Effect>,
         effectHandler: C
     ) -> Builder<Model, Event, Effect> where C.InputType == Effect, C.OutputType == Event {
         return Builder(
@@ -63,7 +73,7 @@ public extension Mobius {
         private let eventConsumerTransformer: ConsumerTransformer<Event>
 
         fileprivate init<C: Connectable>(
-            update: @escaping Update<Model, Event, Effect>,
+            update: Update<Model, Event, Effect>,
             effectHandler: C,
             initiator: @escaping Initiator<Model, Effect>,
             eventSource: AnyEventSource<Event>,
@@ -204,22 +214,13 @@ class LoggingInitiator<Model, Effect> {
     }
 }
 
-class LoggingUpdate<Model, Event, Effect> {
-    private let realUpdate: Update<Model, Event, Effect>
-    private let willUpdate: (Model, Event) -> Void
-    private let didUpdate: (Model, Event, Next<Model, Effect>) -> Void
-
-    init<L: MobiusLogger>(_ realUpdate: @escaping Update<Model, Event, Effect>, _ logger: L) where L.Model == Model, L.Event == Event, L.Effect == Effect {
-        self.realUpdate = realUpdate
-        willUpdate = logger.willUpdate
-        didUpdate = logger.didUpdate
-    }
-
-    func update(_ model: Model, _ event: Event) -> Next<Model, Effect> {
-        willUpdate(model, event)
-        let result = realUpdate(model, event)
-        didUpdate(model, event, result)
-
-        return result
+extension Update {
+    func logging<L: MobiusLogger>(_ logger: L) -> Update where L.Model == Model, L.Event == Event, L.Effect == Effect {
+        return Update { model, event in
+            logger.willUpdate(model: model, event: event)
+            let next = self.update(model: model, event: event)
+            logger.didUpdate(model: model, event: event, next: next)
+            return next
+        }
     }
 }
--- a/MobiusCore/Source/Mobius.swift
+++ b/MobiusCore/Source/Mobius.swift
@@ -19,10 +19,11 @@
 
 import Foundation
 
+/// A wrapper around an update function.
 public struct Update<Model, Event, Effect> {
     private let update: (Model, Event) -> Next<Model, Effect>
 
-    public init(update: @escaping (Model, Event) -> Next<Model, Effect>) {
+    public init(_ update: @escaping (Model, Event) -> Next<Model, Effect>) {
         self.update = update
     }
 
--- a/MobiusCore/Source/EventProcessor.swift
+++ b/MobiusCore/Source/EventProcessor.swift
@@ -67,14 +67,9 @@ class EventProcessor<Model, Event, Effect>: Disposable, CustomDebugStringConvert
 
     func accept(_ event: Event) {
         access.guard {
-            if let current = self.currentModel {
-                let next = self.update.update(model: current, event: event)
-
-                if let newModel = next.model {
-                    self.currentModel = newModel
-                }
-
-                self.publisher.post(next)
+            if self.currentModel != nil {
+                let effects = self.update.update(into: &self.currentModel!, event: event)
+                self.publisher.post(.next(self.currentModel!, effects: effects))
             } else {
                 self.queuedEvents.append(event)
             }
--- a/MobiusCore/Source/EventProcessor.swift
+++ b/MobiusCore/Source/EventProcessor.swift
@@ -42,7 +42,7 @@ class EventProcessor<Model, Event, Effect>: Disposable, CustomDebugStringConvert
     }
 
     init(
-        update: @escaping Update<Model, Event, Effect>,
+        update: Update<Model, Event, Effect>,
         publisher: ConnectablePublisher<Next<Model, Effect>>,
         accessGuard: ConcurrentAccessDetector = ConcurrentAccessDetector()
     ) {
@@ -68,7 +68,7 @@ class EventProcessor<Model, Event, Effect>: Disposable, CustomDebugStringConvert
     func accept(_ event: Event) {
         access.guard {
             if let current = self.currentModel {
-                let next = self.update(current, event)
+                let next = self.update.update(model: current, event: event)
 
                 if let newModel = next.model {
                     self.currentModel = newModel
--- a/MobiusCore/Source/LoggingAdaptors.swift
+++ b/MobiusCore/Source/LoggingAdaptors.swift
@@ -59,9 +59,9 @@ extension Update {
     func logging<L: MobiusLogger>(_ logger: L) -> Update where L.Model == Model, L.Event == Event, L.Effect == Effect {
         return Update { model, event in
             logger.willUpdate(model: model, event: event)
-            let next = self.update(model: model, event: event)
-            logger.didUpdate(model: model, event: event, next: next)
-            return next
+            let effects = self.update(into: &model, event: event)
+            logger.didUpdate(model: model, event: event, next: Next(model: model, effects: effects))
+            return effects
         }
     }
 }
--- a/MobiusCore/Source/MobiusLoop.swift
+++ b/MobiusCore/Source/MobiusLoop.swift
@@ -104,7 +104,7 @@ public final class MobiusLoop<Model, Event, Effect>: Disposable, CustomDebugStri
 
     // swiftlint:disable:next function_parameter_count
     static func createLoop<C: Connectable>(
-        update: @escaping Update<Model, Event, Effect>,
+        update: Update<Model, Event, Effect>,
         effectHandler: C,
         initialModel: Model,
         initiator: @escaping Initiator<Model, Effect>,
@@ -114,7 +114,7 @@ public final class MobiusLoop<Model, Event, Effect>: Disposable, CustomDebugStri
     ) -> MobiusLoop where C.InputType == Effect, C.OutputType == Event {
         let accessGuard = ConcurrentAccessDetector()
         let loggingInitiator = LoggingInitiator(initiator, logger)
-        let loggingUpdate = LoggingUpdate(update, logger)
+        let loggingUpdate = update.logging(logger)
         let workBag = WorkBag(accessGuard: accessGuard)
 
         // create somewhere for the event processor to push nexts to; later, we'll observe these nexts and
@@ -123,7 +123,7 @@ public final class MobiusLoop<Model, Event, Effect>: Disposable, CustomDebugStri
 
         // event processor: process events, publish Next:s, retain current model
         let eventProcessor = EventProcessor(
-            update: loggingUpdate.update,
+            update: loggingUpdate,
             publisher: nextPublisher,
             accessGuard: accessGuard
         )
--- a/MobiusCore/Test/EventProcessorTests.swift
+++ b/MobiusCore/Test/EventProcessorTests.swift
@@ -118,6 +118,7 @@ class EventProcessorTests: QuickSpec {
     }
 
     let testUpdate = Update<Int, Int, Int> { model, event in
-        Next.next(model + event)
+        model += event
+        return []
     }
 }
--- a/MobiusCore/Test/EventProcessorTests.swift
+++ b/MobiusCore/Test/EventProcessorTests.swift
@@ -117,7 +117,7 @@ class EventProcessorTests: QuickSpec {
         }
     }
 
-    func testUpdate(model: Int, event: Int) -> Next<Int, Int> {
+    let testUpdate = Update<Int, Int, Int> { model, event in
         return Next.next(model + event)
     }
 }
--- a/MobiusCore/Test/EventProcessorTests.swift
+++ b/MobiusCore/Test/EventProcessorTests.swift
@@ -118,6 +118,6 @@ class EventProcessorTests: QuickSpec {
     }
 
     let testUpdate = Update<Int, Int, Int> { model, event in
-        return Next.next(model + event)
+        Next.next(model + event)
     }
 }
--- a/MobiusCore/Test/EventRouterDisposalLogicalRaceRegressionTest.swift
+++ b/MobiusCore/Test/EventRouterDisposalLogicalRaceRegressionTest.swift
@@ -69,9 +69,7 @@ class EventRouterDisposalLogicalRaceRegressionTest: QuickSpec {
                     errorThrown = true
                 })
 
-                let update = Update<Model, Event, Effect> { _, _ in
-                    .dispatchEffects([.effect1])
-                }
+                let update = Update<Model, Event, Effect> { _, _ in [.effect1] }
 
                 controller = Mobius.loop(update: update, effectHandler: effectHandler)
                     .withEventSource(eventSource)
--- a/MobiusCore/Test/EventRouterDisposalLogicalRaceRegressionTest.swift
+++ b/MobiusCore/Test/EventRouterDisposalLogicalRaceRegressionTest.swift
@@ -69,7 +69,7 @@ class EventRouterDisposalLogicalRaceRegressionTest: QuickSpec {
                     errorThrown = true
                 })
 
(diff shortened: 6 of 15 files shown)


MobiusCore/Source/Mobius.swift
```diff
@@ -19,7 +19,26 @@
 
 import Foundation
 
-public typealias Update<Model, Event, Effect> = (Model, Event) -> Next<Model, Effect>
+/// A wrapper around an update function.
+public struct Update<Model, Event, Effect> {
+    private let update: (inout Model, Event) -> [Effect]
```
mbrandonw: This is the crux of the change.

MobiusCore/Source/Mobius.swift
```diff
@@ -19,7 +19,26 @@
...
+
+    public init(_ update: @escaping (inout Model, Event) -> [Effect]) {
+        self.update = update
+    }
+
+    public static func create(_ update: @escaping (Model, Event) -> Next<Model, Effect>) -> Update {
```
mbrandonw: This is to have some form of backwards compatibility. Existing update functions can wrap a signature of the old style by just doing `Update.create { model, event in ... }`. We cannot simply overload the initializer of `Update` to use both styles of signature because it will lead to some ambiguous type checking errors.

rastersize:
There’s a fourth way as well, which there’s already code for in MobiusExtra:

```swift
// Given that MyModel conforms to Copyable (no implementation needed)

func update(model: MyModel, event: Event) -> Next<Model, Effect> {
  return Next(model.copy {
    $0.isAdmin = true
  })
}
```

I’m personally not  sold on this idea of an update function with an inout model parameter. To me it seems like it would bring three downsides compared to my example above:

- Breaking the update function up into sub-functions (one per event, which is a common practice) would be less ergonomic in my opinion. As we’d need to continue with the inout all the way down.
- Testing the code would be less ergonomic as well and harder to reason about. As the test would no longer be a simple input -> output. Instead it would have side effects.
- The flow of the code doesn’t read well to me. To me it looks really weird to have a function both mutate some externally provided state and also return an object.

pettermahlen:
I like the `Copyable` idea. While I think `inout` should be safe in general, it's something that is a little harder to reason about than simple pure functions - for instance, it opens up for action-at-a-distance, if the 'final' `var` is also passed as `inout` to something else that might be executed in a different thread. We can more or less protect against that since Mobius should be owning the reference to the current model, but it would be possible to leak it from an update function.

Are we sure about the performance implications, also? In https://chris.eidhof.nl/post/structs-and-mutation-in-swift/ (which may not be correct any more?), he states that:

> In functions, we can mark parameters as inout. Just like with a regular parameter, a copy of the value gets passed in to the function. However, we can change the copy (it’s as if it were defined as a var). And once the function returns, the original value gets overwritten

Anyway, performance doesn't seem like it's a big concern, for most cases. It seems like you can usually work around scenarios where you seem to want very large models, based on previous analysis we have done.

I think we could get rid of `Next` anyway by using a tuple if we feel like `noChange` is a problem. The reason for having it (instead of just returning the old model and no effects) is that it gives the update function control of whether observers will be notified or not. Returning the old model and no effects leads to a notification of the observers, whereas `noChange` should not.

JensAyton:
I actually implemented this change in October. I then updated every Mobius loop in our main app to use the `inout` style, including changing to mutable models as needed, and most of them were no clearer or simpler afterwards, so I abandoned the direction.

I don’t support making this the preferred way without a clear demonstration that real-world loops are markedly improved by it.

@pettermahlen :
> I like the `Copyable` idea. While I think `inout` should be safe in general, it's something that is a little harder to reason about than simple pure functions - for instance, it opens up for action-at-a-distance, if the 'final' `var` is also passed as `inout` to something else that might be executed in a different thread. We can more or less protect against that since Mobius should be owning the reference to the current model, but it would be possible to leak it from an update function.

If I understand your objection correctly, it’s based on false assumptions. an `inout` variable is semantically equivalent to taking the original value as an argument and returning the new value. It isn’t a pointer and can’t be leaked, and the write conceptually happens exactly once at the end of the invocation. (The lack of a memory model means this only needs to be true to the extent it can be observed on a single thread, so unguarded sequential access could expose intermediate values, but that would be a data race anyway.)

> Are we sure about the performance implications, also?

My understanding would indicate that Brandon is mistaken about the performance angle; copying twice is the default behaviour, and it can only be optimized away when inlining.

mbrandonw:
Thanks @rasterize, @pettermahlen and @JensAyton for the comments! You all bring up very good points. I'll try to respond to some of them to the best of my ability. It seems there were 3 main questions: safety, performance and ergonomics.

#### Safety

In Swift, `inout` is essentially special syntax for a common pattern in programming. Many times functions of the following form naturally appear:

```swift
(S) -> (S, A)
```

Random number generators are exactly this shape:

```swift
struct Gen<A> {
  let run: (Rng) -> (Rng, A)
}
```

where `Rng` represents a way to generate a random `Int`. Each time you run the generator it produces a new `Rng` from which new random values can be computed.

Also parsers are of this shape:

```swift
struct Parser<A> {
  let run: (String) -> (rest: String, match: A?)
}
```

Here you parse a bit of the string to produce an (optional) `A` value and then are left with the rest of the string to parse.

And then of course, we encounter this shape with reducers:

```swift
struct Update<Model, Event, Effect> {
  let run: (Model, Event) -> (Model, [Effect])
}
```

This pattern shows up so often that Swift gives a special syntax for making it more ergonomic and more performant:

```swift
struct Gen<A> {
  let run: (inout Rng) -> A
}

struct Parser<A> {
  let run: (inout String) -> A?
}

struct Update<Model, Event, Effect> {
  let run: (inout Model, Event) -> [Effect]
}
```

Essentially, if a type parameter appears on both the left and right side of a function arrow `->` you can instead make the input `inout` and remove the output.

It's worth nothing that Swift's random number API in the standard library is even designed this way:

```swift
var rng = SystemRandomNumberGenerator()
Int.random(in: 0...10, using: &rng)
Int.random(in: 0...10, using: &rng)
Int.random(in: 0...10, using: &rng)
```

whereas the alternative without `inout` would be:

```swift
var _rng = SystemRandomNumberGenerator()
var result: Int
(result, rng) = Int.random(in: 0...10, using: rng)
(result, rng) = Int.random(in: 0...10, using: rng)
(result, rng) = Int.random(in: 0...10, using: rng)
```

It is also worth noting that the Swift standard library comes with a version of `reduce` that specifically works for `inout`, and the main reason it was [introduced](https://forums.swift.org/t/reduce-with-inout/4897) was for ergonomics and performance:

```swift
[1, 2, 3, 4].reduce(into: 0, +=)
[1, 2, 3, 4].reduce(0, +)
```

`inout` has less to do with "side effects" and mutation, and more with providing a nicer way to perform a very common pattern. And Swift makes this pattern very safe by having compiler guarantees that the `inout` value is not allowed to travel further than the current scope. For example, you cannot capture an `inout` value in some asynchronous work and then try to mutate it later:

```swift
func foo(value: inout Int) {
  value = 1 // ✅ OK
  DispatchQueue.main.asyncAfter(deadline: .now() + 1) {
    value = 2 // 🛑 Escaping closure captures 'inout' parameter 'value'
  }
}
```

This is a compiler error, not a runtime error. You are simply not allowed to pass `inout` values across closure boundaries in Swift.

So, it is very safe to use `inout` and very different from mutation of references. It is just as safe as returning a whole new value, and it makes the function no less "pure".

#### Performance

As for performance, it is indeed a measurable thing and is the motivating reason for its existence and motivates the design of certain APIs. Here are some very simple benchmarks that when run (with optimizations on) show there is a difference in using `inout` and not:

https://gist.github.com/mbrandonw/9f8581f35e173b7acd789c85b39070e5

The optimizations that `inout` can take advantage of are also [documented](https://docs.swift.org/swift-book/ReferenceManual/Declarations.html#ID545). This describes that even though sematically we can think of `inout` as doing a copy and reassignment, in practice the copy can be optimized away for some situations.

So the only question is whether or not this will affect a Mobius feature. As it stands today it may not, but there are maybe two situations where it could be a problem in the future:

* If many Mobius features are composed into a single feature the model can become quite large.
* If a Mobius feature sends lots of events (e.g. timers) the unnecessary copying can start to become a problem.

#### Ergonomics

Then there's also the question of whether or not this is more ergonomic. 

While the `Copyable` protocol gets the job done, it doesn't seem ideal for Mobius to vend this functionality. Users of Mobius.swift may have their own `Copyable` library that will conflict with this one, and that could prevent people from using both libraries together. Also, every model that wants to take advantage of `Copyable` functionality needs to import MobiusCore and MobiusExtras. Ideally this functionality would be split into its own module. But then we should question whether it pulls its own weight over using Swift's well-defined value semantics.

@stephencelis and I have worked on converting many update functions to the `inout` style, and have seen lots of benefits in doing it. It would be easiest to discuss this IRL, but here is a distilled example of a pattern we see a lot.

The following is a real-world example (names have been changed) that wants to change the title of a "project" inside a "model". It needs to do some additional work to mutate the value in `model.project.name`, and the `withXYZ` pattern doesn't help here:

```swift
func didEditTitle(_ model: Group, title: String?) -> Next {
  guard let title = title else {
    return .noChange
  }

  var project = model.project
  project.name = title

  return .next(model.withChanged(group: group))
}
```

Alternatively, if we had `Copyable` at our disposal this could be made simpler:

```swift
func didEditTitle(_ model: Group, title: String?) -> Next {
  if let title = title {
    return .next(model.copy { $0.name = title })
  }
  return .noChange
}
```

But this style isn't that much different from using `inout` directly:

```swift
func didEditTitle(_ model: inout Group, title: String?) -> Next {
  if let title = title {
    model.project.name = title
  }
  return []
}
```

We have a lot more examples, but it might be nice to run through them in person to show how it could work in practice.

---

Thanks again for all of the feedback! We're happy to close this out if it does not feel like its in the spirit of what Mobius.swift is trying to accomplish. We just wanted to float this idea and see if it might help people working on larger Mobius features.

erikprice:
> @stephencelis and I have worked on converting many update functions to the inout style, and have seen lots of benefits in doing it.

I'd be interested in hearing more about these benefits.

> We're happy to close this out if it does not feel like its in the spirit of what Mobius.swift is trying to accomplish.

I hope you don't close this out until we've heard more arguments or evidence for or against. Whether the model parameter is declared `inout` or has its type re-declared in the return seems to have little to do with the spirit of Mobius, as the two approaches are completely isomorphic. And I'm not sure that the performance difference alone would be worth changing the API. So it really boils down to style/ergonomics – which of these is "idiomatic Swift" (an admittedly still-evolving notion), and how each approach affects the call sites.

(My own gut feeling when I first read this PR was that, given Swift has a language-level feature for functions to indicate that they return a possibly-changed copy of their parameters, we would have been using that feature from the beginning if we had thought of using it before. It is the whole point of `inout`. But my gut also still hasn't really accepted the perceived need to make Mobius safe for use from any thread, so take whatever it says with a grain of salt.)

pettermahlen:
I agree with @erikprice, let's try to make sure we agree on the implications of this before deciding either way.

It's very unsurprising that the naive example without inout is slow, as it does 100,000 allocations of arrays of on average 50k elements. It does display how call-by-reference using inout helps with performance, but it's also not likely to be something that'll impact any Mobius loop. When creating the Java version, we explicitly decided against targeting high-volume, low-latency performance. Not sure if we documented that very well but the point was that designing for high performance is very hard and our focus was to simplify business logic where you need to process  at most a couple of events per second. So I don't think performance is a very strong argument either way.

Regarding the safety, as I read https://docs.swift.org/swift-book/ReferenceManual/Declarations.html#ID545, I still worry about safety loopholes in the presence of call-by-reference (I also don't like `var` properties in structs as I think they muddy the waters with regard to what is actually mutable or not; relying on declaration-site immutability makes me uncomfortable). I can see that that worry might not be motivated if the compiler can provide sufficient guarantees.

In terms of ergonomics, it seems to me that there's no real difference between `Copyable` and `inout` - neither allow truly immutable structs (with `let` properties) and both lead to about the same amount of boilerplate.

Another point we didn't discuss is learnability/approachability. I would argue `inout` is not a trivial concept, in particular not if you're someone who wants to understand in detail what the machine will do when it runs into an `inout` parameter. But if you're happy to ignore those details, then I would assume that most Swift developers will have come across the concept and understood how to use it, after some time using the language. `Copyable` is (much) simpler but will be less common and isn't a part of the language. I don't think it would take more than a couple of minutes to learn, though.

Regarding thread-safety, I've been the one most strongly pushing for it. That probably comes from my gut being used to working in multithreaded backend service code, where there is no choice but to make things thread-safe by default. The current changes being worked on in fact make `MobiusLoop` safe only for single-threaded use (arguably, I suppose, but let's not go there). I think as we start using more and more of Combine, it's likely that threading will become more noticeable. In general, of course, to get more performance out of the hardware, it'll be impossible to stick to a single thread going forward. Maybe for the kinds of cases we're targeting, that level of performance won't be necessary, and so we can get away with only supporting single-threaded use. But I'm pretty sure we need to have at least well-defined ways of using Mobius in a safe way from multiple threads and queues, even if they are less fool-proof than the guarantees we provide in the Java version.

I'm not sure about this. I don't like `inout` because I don't think it's simple - it is most definitely mutation in-place, even to the point of using references/pointers and therefore inherently unsafe and more complex than values in, values out. But I think the strong argument in favour of it is that it's what Apple recommends and therefore what I think most Swift developers would expect to use. Would be great with a time machine that could tell us what the consensus about `inout` is 5-10 years into the future. :)

jeppes:
I largely agree with @pettermahlen and @JensAyton on this.

If a compelling enough case could be made for performance, I think it would be reasonable to use this for the internal representation. But I don't think that case has been made.

I don't think the argument that the two approaches are isomorphic is very relevant. Swapping out isomorphisms is not something which should be done trivially. As @pettermahlen has said, the 'values in, values out' approach is fundamentally simpler, and encourages you to think of the update function as a something which produces a new state based on some knowledge of the past. On the other hand, `inout` mutation makes it seem like the past is being overwritten. I think it's much easier to reason about a loop as a series of immutable models instead of one model which is different depending on when you choose to look at it. Even if `inout` is safe, it's still fundamentally more complicated, and understanding how and when it works (which I think is required to use it effectively) is unnecessarily difficult.

As for the `Rng` example mentioned in a comment: I don't think this is very compelling. It shows that using `inout` sets up implicit dependencies between the lines of code. Even if this mutation is local, it requires knowledge about the entire scope to reason about the effect of any individual line using the `rng` variable. When a bug inevitably appears in the program, understanding the entire scope of this mutation (which can be greatly expanded by passing it around as an `inout`) will be required. This runs the risk of becoming significantly more difficult if the type in question happens to either be a reference-type, reference a reference-type, or be referenced by a reference-type.

erikprice:
🤦‍♂ The whole point of mentioning that `inout` is isomorphic to `A -> A` is to highlight the fact that neither is "fundamentally simpler" than the other. One could equally argue that the presence of `inout` on the function parameter "encourages you to think of the update function as something which produces a new state based on some knowledge of the past".

*In actual practice* let's keep something in mind: _most_ users of Mobius.swift are going to be _authoring_ update functions, not _calling_ them (except in tests of course). For someone writing an update function, which of these two approaches offers a bigger pit of success?

Changing the subject: earlier today I was wondering whether introducing the `inout` approach would have any impact on the ability to compose update functions – I suppose not, but just want to make sure I'm not missing something. (Also, maybe update function composition isn't really that important if we're encouraging composing Mobius loops now?)

jeppes:
> One could equally argue that the presence...

No, `inout` does not encourage that style of thinking: The whole point of it is to enable in-place mutation.

> The whole point of mentioning that inout is isomorphic to A -> A is to highlight the fact that neither is "fundamentally simpler" than the other.

The two things being isomorphic has nothing to do with their simplicity. Adding `inout` couples the function to the scope in which it is called. Such a function is therefore fundamentally more complicated than a function which is independent of the scope in which it is called.

pettermahlen:
> > The whole point of mentioning that inout is isomorphic to A -> A is to highlight the fact that neither is "fundamentally simpler" than the other.
> 
> The two things being isomorphic has nothing to do with their simplicity. Adding `inout` couples the function to the scope in which it is called. Such a function is therefore fundamentally more complicated than a function which is independent of the scope in which it is called.

I think this hits the nail on the head - the problem with `inout` is that it couples the function to the calling scope. (Similar with `var` properties in structs, which means the behaviour depends on how the variable holding the struct was declared). This improves 'writability' at the cost of degrading simplicity/understandability. Having said that, in the context of a Mobius update function, we *have* a very well-defined scope for the function.

I really don't know which way to turn on this - in almost all cases, I think it's clearly a better choice to prioritise even a small improvement in understandability over an improvement in 'writability'. You read code about 5-10 times more than you write it, according to research. In this case, since things are so well controlled, perhaps that heuristic doesn't apply. On the other hand, maybe we should discourage the use of `inout` because it's inherently complex? That seems like an uphill battle and maybe Apple can provide guarantees that mean you don't need to understand the underlying complexity and can use it completely safely, treating it a little like magic.

JensAyton:
I don’t buy the “couples to scope” argument. All the `inout` tells you _as the callee_ is that the caller will provide a value of type `T` and consume a value of the same type `T`. That’s exactly the same degree of coupling as a `(T) -> T`.

pettermahlen:
No, because `inout` means you have to make an in-place update of something that is therefore by necessity mutable. This means you open up for action-at-distance. The `inout` thing your function is mutating might be changed by someone else as well. There is underlying complexity there - which as I said earlier, Apple might successfully hide to the extent you normally don't need to worry about it, but it's still there and a difference between a true pure function working with immutable values (is that a tautology?).

JensAyton:
The risk of someone else changing the value does not exist, or rather, does not exist any more than it does for an arbitrary parameter. A storage location captured with `&` for use with `inout` is considered exclusively accessed; it cannot legally be read or written to except through the `inout`. Semantically, it will be read from exactly once at the beginning of the access, and written to exactly once at the end of the access. In this sense, `var x = ...; foo(&x)` is precisely equivalent to `var x = ...; x = foo(x)`. (The cross-module calling convention for an inout parameter consists of a normal in parameter plus a second one passing a closure to write back the result.)

When the optimiser can statically guarantee that exclusive access rules are being followed, and a call is happening within the same linkage unit, it’s free to replace the standard calling convention for `inout` with a borrowed reference, i.e. a non-refcounting reference for classes and a pointer to storage for other types. This would potentially allow the storage to be overwritten by code where exclusive access isn’t enforced, e.g. via an unsafe pointer, C code, or from another thread. However, _exactly the same is true for normal arguments_, since they can also be promoted to borrows.

There is one abstraction leak: copy-on-write. When a normal argument is turned into a borrow, refcounting is elided and this can be observed with `isKnownUniquelyReferenced`. This is the performance benefit mentioned upthread; by passing a collection `inout` it can be mutated without copying, in cases where this is safe, and the call is transparent to the optimizer. The optimizer apparently can’t make this kind of transformation on `(T) -> T` functions.

Aside from the esoteric possibility of a class-based model using `isKnownUniquelyReferenced` in an unintended way, `(inout T) -> Void` and `(T) -> T` are equivalent and neither implies anything about the calling context.

JensAyton:
Sorry, I lost focus and the bit about optimizations is inaccurate. `(T) -> T` functions obviously don’t break copy-on-write all the time or it would be largely pointless.

The actual optimization opportunity is code like `foo(&someDictionary[someKey])`. In this case, instead of a `get` access and a `set` access, a single `_modify` coroutine-based access is used,  and the item from the dictionary is borrowed. If that item is itself a COW container, this can avoid unnecessary copying.

This works because the `inout` has exclusive access to `someDictionary` for the duration of the execution of `foo`. In the non-`inout` case `someDictionary[someKey] = foo(someDictionary[someKey])`, there are two different accesses and the dictionary _can_ be modified between them.

This is true of simple variables as well, so I’ll modify my conclusion: the difference between `(inout T) -> Void` and `(T) -> T` is that the `inout` variant guarantees that the captured location will not be modified by other means, while the simple function does not.

mbrandonw:
> This means you open up for action-at-distance.

We may all need to be on the same page of what is meant by "action-at-distance", because what `inout` is doing quite a bit different from what is typically thought of as "action at a distance".

Here is a pretty standard example of "action at a distance":

```swift
class CounterClass { var count = 0 }

func handle(counter: CounterClass) {
    DispatchQueue.main.asyncAfter(deadline: .now() + 1) {
        counter.count = 1
    }
}

let counter = CounterClass()
handle(counter: counter)
print(counter.count) // 0
```

The "distance" here is the fact that we give our counter to the `handle` function, and immediately after we can observe that nothing about the counter has changed. Yet a second later its count will be magically changed to 1, and even worse, _everyone_ that holds a reference to `counter` will also see this magical, distant mutation and no one will know where it came from. There is nothing in the function signature that let's us know this is going to happen.

The equivalent Swift code for structs will not compile, and is in fact impossible to implement:

```swift
struct CounterStruct { var count = 0 }

func handle(counter: inout CounterStruct) {
    DispatchQueue.main.asyncAfter(deadline: .now() + 1) {
        counter.count = 1 // 🛑 Escaping closure captures 'inout' parameter 'counter'
    }
}

var counter = CounterStruct()
handle(counter: &counter)
```

Swift simply will not allow a value type to be mutated across this asynchronous boundary.

The furthest distance one can get `inout` mutations to travel is limited to the current scope, and special syntax is even need to make it very explicit (`&`):

```swift
func handle(value: inout Int) { value += 1 }

var x = 1
handle(value: &x)
print(x)
```

But this distance is no further than the following:

```swift
var x = 1
x = 2
print(x)
```

which I don't think classifies as what is typically known as "action at a distance".

pettermahlen:
Yes... I am reading what both of you saying as arguments in favour of accepting this because as I said earlier, the complexities inherent in `inout` are hidden behind walls that force users to use it safely, and therefore it'll be fine. There's no doubt in my mind that `inout` is more 'easy' than 'simple' as Rich Hickey defines it, and I am philosophically opposed to making mutation the 'default mode of operation', in the way that `inout` does. But I think I should defer to your arguments and Swift expertise so I guess since you're in favour of this way of writing update functions, I should not block it.

I'd be surprised if `inout` is seen as a great thing that will be adopted by other languages too, but I'm wrong about stuff all the time. :)

pettermahlen:
@mbrandonw could you update the PR so it builds?

JensAyton:
In Rust, it’s spelled `&mut`. 🙃 If your language wants mutability, exclusive access guarantees, and efficiency, it seems to fall out naturally.

Meta-semantics aside, I stand by my previous comment:
> I actually implemented this change in October. I then updated every Mobius loop in our main app to use the `inout` style, including changing to mutable models as needed, and most of them were no clearer or simpler afterwards, so I abandoned the direction.
>
> I don’t support making this the preferred way without a clear demonstration that real-world loops are markedly improved by it.

stephencelis:
@JensAyton Your comment was the one that sent us down this path to PR 😄: https://github.com/spotify/Mobius.swift/pull/92#discussion_r358205766

stephencelis:
@JensAyton Perhaps it'd be good to open up the discussion a bit wider. A team we are working with ended up with a bug that was difficult to troubleshoot because in constructing a brand new value in `.next`, they were trampling over another value with a default. Their solution to avoid the bug ended up first introducing the `var model = model` dance in order to mutate the current value, and then introducing a `copy` helper function that hid the dance away, much like the `Copyable` protocol in MobiusExtras. Supporting Swift's `inout` semantics by default would avoid some of these bugs and reinventions while allowing folks to delete and maintain less code in the process.

JensAyton:
@stephencelis I think this points at something we can surely agree on: manually written field-by-field copying functions are not only annoying boilerplate but a completely avoidable source of bugs.

If we end up not supporting `inout`, I’ll certainly advocate a more focussed copy-by-mutation method like `MobiusExtras.Copyable`. Possibly that approach would be more palatable to @pettermahlen and @jeppes, since the mutation is more closely contained; so far I’ve found it to be a pretty nice option stylistically as well.

JensAyton:
We discussed this today and came to the following conclusion:

The advocates of immutability are willing to accept that `inout` is idiomatic Swift for this kind of local state change, and doesn’t add risks. However, on the whole we are not convinced that the `inout` signature necessarily produces more legible code in practice than a `MobiusExtras.Copyable`-style helper or that a change to the core definition is warranted.

We recommend that you continue your prototyping using either `MobiusExtras.Copyable` or an extension method on `Mobius.Builder`. A real-world code base using `inout` mutation style should provide a better basis for deciding whether to make this change in the future.

JensAyton:
N.b. the above was written without regard for #92, and with that in place an `inout` function as an alternative form for the `Update` struct may be better received, we didn’t actually discuss the combination.

PR #94, opened 2019-12-18
Replace struct EffectHandler with protocol

jeppes:
Based on feedback from the new EffectHandler changes, it seems reasonable to replace the `EffectHandler` struct with a protocol. This will make it easier to create effect handling classes.

I think the main issue here is how to handle cancellation.

1:
``` swift
protocol EffectHandler {
    func handle(_ input: Input, _ output: @escaping Consumer<Output>) -> Disposable
}
```
2:
```swift
protocol EffectHandler {
    func handle(_ input: Input, _ output: @escaping Consumer<Output>)

    func dispose()
}
```

1 has the advantage of fine-grained control over disposal. It also enables effect handlers to be implemented as structs. The issue, however, is that it requires a policy for how to deal with previously created disposables. The current solution is to gather all the created disposables and dispose them all when the loop is torn down.
Please let me know what you think.

@JensAyton @pettermahlen


MobiusCore/Source/EffectHandlers/EffectHandler.swift
```diff
@@ -17,16 +17,23 @@
...
-    public let disposable: Disposable
+public protocol EffectHandler {
+    associatedtype Input
+    associatedtype Output
+
+    func handle(_ input: Input, _ output: @escaping Consumer<Output>) -> Disposable
+}
+
+public struct AnyEffectHandler<Input, Output>: EffectHandler {
+    private let handler: (Input, @escaping (Output) -> Void) -> Disposable
```
pettermahlen: What is the rationale for not using the `Consumer` type alias here? It seems like using it would make the code a little clearer, because you get a hint about what the closure is for.

pettermahlen: (The same comment applies to other places in this file)

MobiusCore/Source/EffectHandlers/EffectRouter.swift
```diff
@@ -58,19 +60,23 @@ private struct Route<Input, Output> {
...
     let disposable: Disposable
 
-    init<Payload>(
+    init<Payload, Handler: EffectHandler>(
         path tryPath: @escaping (Input) -> Payload?,
-        handler: EffectHandler<Payload, Output>
-    ) {
+        handler: Handler
+    ) where Handler.Input == Payload, Handler.Output == Output {
+        var disposables: [Disposable] = []
```
pettermahlen: Does not this variable scope mean that two concurrent invocations of the effect handler use the same array instance? Not necessarily concurrent on different threads; I guess depending on the semantics of handling an effect/sending events, the effect handler could be recursively invoked before it completes. Maybe that is safe. 

Also, I think the array will grow indefinitely? It seems like there needs to be a way to remove disposables once there is no longer a need for cancellation? Is it possible to write a test that verifies that there is no memory leak?

MobiusCore/Source/EffectHandlers/EffectHandler.swift
jeppes: No reason, I just forgot

MobiusCore/Source/EffectHandlers/EffectRouter.swift
jeppes: The access to the blocks which modify this array is synchronized, so it should not be an issue.

The garbage thing is an issue though. I will write about it in a comment in the main PR thread

jeppes:
@pettermahlen I definitely agree with all your points here.
The solution as proposed here is not sufficient.

If there is a way to make the `handle(..) -> Disposable` solution work, I think it would be really beneficial. I will devote more time to thinking about it. Any ideas or input would be great.

jeppes:
The issue is that we need to know when an effect has finished:
I have one idea which is somewhat Rx inspired:

```swift
EffectRouter()
    .routeEffects(...).to { effect, output in
        let request = http.request(...) { result
            output.send(Event.result(result))
            output.end()
        }
        return request.cancel
    }
```
where `.to` is
```swift
func to(effect: Effect, output: Output<Event>) -> Disposable
```
where `Output` is
```swift
protocol Output<T> {
    func send(_: T)
    func end()
}
```

We could (and should) wrap up the argument to `.to` in a type, so that classes/structs can be created which implement it:
```swift
.to(EffectHandler)

protocol EffectHandler<Effect, Event> {
    func handle(effect: Effect, output: Output<Event>) -> Disposable
}
```

------------------------------------------

The downside here is that the user would need to know that they should call `end`. It would likely not be a big deal if `end` was not called in most cases, but not doing so in a long running process would cause memory leaks. The user would also have to know not to call `send` after `end`: I don't think this is a very big deal. Any violation of it would likely be caught early, and could result in an exception in debug builds.

Another possible approach would be to only have one method: `end(_: Event)`, which would signal termination with some output. I like this approach less, since it would require output in all cases, and it would not support multiple events to be emitted for a single effect.

Thoughts? @pettermahlen

jeppes:
After a discussion in person, @pettermahlen and I have agreed on the design of the `Output` protocol. Although we both agree that the name should be changed.
A common design for these interfaces is to include an `error(_: Error)` callback. We have decided against this, preferring instead to propagate errors back into the loop with events, so that error handling is encouraged to be baked into the loop logic.

JensAyton:
Did this discussion in person consider and reject the option of `end([T])` (or some other name) as the only method? It addresses both of the concerns originally raised.

MobiusCore/Source/EffectHandlers/EffectHandler.swift
```diff
@@ -17,16 +17,44 @@
...
+    /// When you are done handling `input`, be sure to call `response.end()` to prevent memory leaks.
+    ///
+    /// Return a `Disposable` which tears down any resources that is being used by this effect handler.
+    func handle(
+        _ input: Effect,
+        _ response: Response<Event>
+    ) -> Disposable
+}
+
+public class Response<T> {
```
JensAyton: `final`?

MobiusCore/Source/EffectHandlers/EffectHandler.swift
```diff
@@ -17,16 +17,44 @@
...
-        disposable: Disposable
+        onSend: @escaping (T) -> Void,
+        onEnd: @escaping () -> Void
     ) {
-        self.handle = handle
-        self.disposable = disposable
+        self.send = onSend
+        self.end = onEnd
+    }
+}
```
JensAyton: I would much prefer it if `Response` called `end` in `deinit`. If that’s not acceptable, it should `fatalError` if `end` hasn’t been called. I don’t think silently leaking responses would be acceptable.

MobiusCore/Source/EffectHandlers/EffectHandler.swift
```diff
@@ -17,16 +17,44 @@
...
-    public let disposable: Disposable
+public protocol EffectHandler {
+    associatedtype Effect
+    associatedtype Event
+
+    /// Handle an `Effect`.
+    /// To output events, call `response.send`
+    /// When you are done handling `input`, be sure to call `response.end()` to prevent memory leaks.
+    ///
+    /// Return a `Disposable` which tears down any resources that is being used by this effect handler.
```
JensAyton: This should clearly document which orderings of `dispose` and `end` are allowed.

jeppes:
@JensAyton Yes, but that's a good point. I forgot to write down the reasoning.

The problem with that is that it does not allow multiple effects to be dispatched over time. You might, for example, want to do something like:
```
.routeEffects(...).to { effect, response in
    response.send(.downloadStarted)
    makeRequest()
        .then { result in
            response.send(.downloadFinished)
            response.end()
        }
    ...
}
```

MobiusCore/Source/EffectHandlers/EffectHandlingConnection.swift
```diff
@@ -0,0 +1,117 @@
...
+        self.handleEffect = handleInput
+        self.unsafeOutput = unsafeOutput
+    }
+
+    func handle(_ effect: Effect) -> Bool {
+        let id = UUID()
+        let response = Response(onSend: self.output, onEnd: { [unowned self] in self.delete(id: id) })
+        let effectHandlerState: EffectHandlingState<Event> = .unhandled(response: response)
+
+        create(id: id, state: effectHandlerState)
```
jeppes: Calling `create` here is costly and technically unnecessary.
Ideally we should only store the state once handleEffect(effect, response) has returned. The problem though, is that `handleEffect` could call `response.end` before it returns. In that case, it should not be stored.

Any other ideas here would be appreciated.

MobiusCore/Source/EffectHandlers/EffectHandlingConnection.swift
```diff
@@ -0,0 +1,117 @@
...
+    init(
+        handleInput: @escaping (Effect, Response<Event>) -> Disposable?,
+        output unsafeOutput: @escaping Consumer<Event>
+    ) {
+        self.handleEffect = handleInput
+        self.unsafeOutput = unsafeOutput
+    }
+
+    func handle(_ effect: Effect) -> Bool {
+        let id = UUID()
```
JensAyton: An incrementing counter would be cheaper. :-)

MobiusCore/Source/EffectHandlers/EffectRouterDSL.swift
```diff
@@ -61,14 +69,12 @@ public extension PartialEffectRouter {
         _ connectable: C
     ) -> EffectRouter<Input, Output> where C.InputType == Payload, C.OutputType == Output {
         var connection: Connection<Payload>?
-        return to(EffectHandler<Payload, Output>(
-            handle: { payload, output in
-                connection = connection ?? connectable.connect(output)
-                connection?.accept(payload)
-            },
-            disposable: AnonymousDisposable {
+        return to { payload, response in
```
jeppes: This will leak memory since `response.end()` is never called, so `EffectHandlingConnection` will accumulate entries for each effect it receives.

MobiusCore/Source/EffectHandlers/EffectHandlingConnection.swift
```diff
@@ -0,0 +1,117 @@
...
+            guard let state = self.handlingEffects[id] else {
+                return
+            }
+            self.handlingEffects[id] = transform(state)
+        }
+    }
+
+    private func delete(id: UUID) {
+        lock.synchronized {
+            _ = self.handlingEffects.removeValue(forKey: id)
```
JensAyton: `self.handlingEffects[id] = nil` would be more idiomatic

MobiusCore/Source/EffectHandlers/EffectHandlingConnection.swift
JensAyton: Sounds like you need to smuggle state on a side channel. :-)

One option would be to make `id` a `var` and clear it in `onEnd`. `create` would then only be called if it’s non-nil after `handleEffect`.

MobiusCore/Source/EffectHandlers/EffectRouterDSL.swift
```diff
@@ -28,31 +28,39 @@ public extension EffectRouter where Input: Equatable {
 }
 
 public extension PartialEffectRouter {
+    /// Route to an `EffectHandler`.
+    /// - Parameter effectHandler: the `EffectHandler` for the route in question.
```
JensAyton: Comment is wrong

MobiusCore/Source/EffectHandlers/EffectRouterDSL.swift
```diff
@@ -28,31 +28,39 @@ public extension EffectRouter where Input: Equatable {
...
+        return to { payload, response in
+            if let event = eventClosure(payload) {
+                response.send(event)
+            }
+            response.end()
+            return AnonymousDisposable {}
+        }
     }
 
     /// Route to a side-effecting closure.
```
JensAyton: This comment is also wrong :-)

MobiusCore/Source/EffectHandlers/EffectRouterDSL.swift
JensAyton: I think the `Response` design conflicts with [backwards compatibility for connectables](https://github.com/spotify/Mobius.swift/pull/95), so this function should be removed.

The alternative design of having one dispose/end per handler rather than per invocation maps cleanly to connectables.

jeppes:
@pettermahlen @JensAyton

MobiusCore/Source/EffectHandlers/EffectExecutor.swift
```diff
@@ -0,0 +1,100 @@
...
+            // Any events produced as a result of handling the effect will be sent to this class's `output` consumer,
+            // unless it has already been disposed.
+            onSend: { [weak self] event in self?.output?(event) },
+            // Once an effect has been handled, remove the reference to its response and disposable.
+            onEnd: { [weak self] in self?.delete(id: id) }
+        )
+
+        let disposable = handleEffect(effect, response)
+
+        // If the effect has not yet been handled, store a reference to its response and disposable.
```
pettermahlen: This is actually racy - it works for the synchronous case, but may break for 'fast async' cases. That is, `Response.end` may be invoked between the check of 'ended' and the call to 'store'. 

I think it's safe to write this instead:

```
store(id: id, response: response, disposable: disposable)
if response.ended {
    delete(id: id)
}
```

Since duplicate deletions should be fine.

MobiusCore/Source/EffectHandlers/EffectExecutor.swift
```diff
@@ -0,0 +1,100 @@
...
+    // When disposing this effect handler, all entries must be removed.
+    private var handlingEffects: [Int64: EffectHandlingState<Event>] = [:]
+    private var nextID = Int64(0)
+
+    init(handleInput: @escaping (Effect, Response<Event>) -> Disposable) {
+        self.handleEffect = handleInput
+    }
+
+    func connect(_ consumer: @escaping Consumer<Event>) -> Connection<Effect> {
+        output = consumer
```
pettermahlen: This should check if we're already connected to something - otherwise, there's a risk that the new consumer gets events intended for the previous one via the `output`.

pettermahlen: (Should be easy to guard with a test, too)

MobiusCore/Source/EffectHandlers/EffectRouter.swift
```diff
@@ -17,6 +17,27 @@
...
+/// The first is the routing criteria. There are two choices here:
+///  - `.routeEffects(equalTo: constant)` - Routing to effects which are equal to `constant`.
+///  - `.routeEffects(withPayload: extractPayload)` - Routing effects that satisfy
+///     a payload extracting function: `(Effect) -> Payload?`. If this function returns a non-`nil` value,
+///     that route is taken and the non-`nil` value is sent as the input to the route.
+///
+/// These two routing criteria can be matched with one of four types of targets:
+///  - `.to { effect in ... }` - A fire-and-forget style function of type `(Effect) -> Void`.
+///  - `.toEvent { effect in ... }` A function which returns an optional event to send back into
+///     the loop: `(Effect) -> Event?`.
```
pettermahlen:
```suggestion
///     the loop: `(Effect) -> Event?`. This makes it easy to send a single event caused by the effect.
```

MobiusCore/Source/EffectHandlers/EffectRouter.swift
```diff
@@ -37,6 +58,10 @@ public struct EffectRouter<Input, Output> {
         return PartialEffectRouter(routes: routes, path: payload)
     }
 
+    /// Convert this `EffectRouter` into `Connectable` which can be attached to a Mobius Loop, or called on its own to handle effects.
+    ///
+    /// Note: Each instance of `Input` must have been linked to exactly one handler. A runtime crash will occur if zero or multiple handlers were found for some
```
pettermahlen: Maybe this note belongs in the top-level documentation? I guess most people won't be invoking `toConnectable` themselves and so might not see this comment?

MobiusCore/Source/EffectHandlers/Response.swift
```diff
@@ -0,0 +1,70 @@
...
+// under the License.
+
+import Foundation
+
+/// A `Response` can send output and signal completion.
+/// Sending output is done with `.send` and signaling completion is done with `.end`.
+///
+/// Note: Once `.end` has been called (from any thread), the closure you provide in `onSend` will no longer be called.
+/// Note: The closure you provide in `onEnd` will only be called once when `.end` is called on this object.
+public final class Response<T> {
```
pettermahlen: It would be useful to harmonise the names used here with the ones used in the Kotlin experiments. Maybe that should just be a note for now, because we're not really yet aiming at creating the final API, I think. For reference, the work in progress there defines an `Effect.Callback` which fills the same role as `Response`, and has the methods `onStart(Disposable)`, `onValue` and `onComplete`. It's also to be noted that it's the same thing as a Reactive Streams [Subscriber](https://github.com/reactive-streams/reactive-streams-jvm#2-subscriber-code).

MobiusCore/Source/EffectHandlers/ThreadSafeConnectable.swift
```diff
@@ -0,0 +1,65 @@
...
+    private var connection: Connection<Effect>?
+
+    init<Conn: Connectable>(
+        connectable: Conn
+    ) where Conn.InputType == Effect, Conn.OutputType == Event {
+        self.connectable = AnyConnectable(connectable)
+    }
+
+    func connect(_ output: @escaping (Event) -> Void) -> Connection<Effect> {
+        lock.synchronized {
```
pettermahlen: Like the `EffectExecutor`, this method should guard against duplicate connection attempts.

MobiusCore/Source/EffectHandlers/ThreadSafeConnectable.swift
```diff
@@ -0,0 +1,65 @@
...
+
+            return Connection(
+                acceptClosure: accept,
+                disposeClosure: dispose
+            )
+        }
+    }
+
+    private func accept(_ effect: Effect) {
+        connection?.accept(effect)
```
pettermahlen: Since the `connection` field may be read and written by different threads, I would assume that every access to it should be guarded by the lock.

MobiusCore/Source/EffectHandlers/ThreadSafeConnectable.swift
pettermahlen: (Including in `dispose` below).

MobiusCore/Source/EffectHandlers/ThreadSafeConnectable.swift
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    private func dispose() {
+        var shouldDispose = false
+        lock.synchronized {
+            shouldDispose = output != nil
+            output = nil
+        }
+        if shouldDispose {
+            connection?.dispose()
```
pettermahlen: Should probably set the connection to nil here, too? In case we want to support the connect/disconnect/connect case. A good test case to write if so.

MobiusCore/Test/EffectHandlers/EffectHandlerDSLTests.swift
```diff
@@ -36,26 +36,89 @@ private enum Event: Equatable {
 class EffectRouterDSLTests: QuickSpec {
```
pettermahlen: Maybe rename this file since it's now EffectRouterDSL?

MobiusCore/Test/EffectHandlers/EffectHandlerDSLTests.swift
```diff
@@ -36,26 +36,89 @@ private enum Event: Equatable {
...
+            }
+
+            it("should not be disposed if end was called") {
+                connection.accept(1)
+                connection.dispose()
+
+                expect(wasDisposed).to(beFalse())
+            }
+
+            it("should be disposed if end was never called") {
```
pettermahlen: The test description doesn't match the implementation. Is it testing the right thing?

MobiusCore/Test/EffectHandlers/EffectHandlerDSLTests.swift
```diff
@@ -36,26 +36,89 @@ private enum Event: Equatable {
...
+        }
+
+        context("An EffectHandler which calls end after some time") {
+            var wasDisposed: Bool!
+            var connection: Connection<Int>!
+            var end: (() -> Void)!
+
+            beforeEach {
+                wasDisposed = false
+                end = {}
```
pettermahlen: Is this initialisation needed? It looks a little weird. I think it would make sense to crash the test if an attempt at calling `end` before it has the right value is made.

MobiusCore/Test/EffectHandlers/EffectHandlerDSLTests.swift
```diff
@@ -36,26 +36,89 @@ private enum Event: Equatable {
...
+        }
+
         context("Effect routers based on constants") {
             it("Supports routing to an effect handler") {
                 var events: [Event] = []
                 var wasDisposed = false
-                let effectHandler = EffectHandler<Effect, Event>(
-                    handle: { effect, dispatch in
+
+                let dslHandler = EffectRouter<Effect, Event>()
```
pettermahlen:
```suggestion
                let connection = EffectRouter<Effect, Event>()
```

MobiusCore/Source/EffectHandlers/EffectRouter.swift
jeppes: Most users would be invoking it. I'll move the comment

MobiusCore/Source/EffectHandlers/Response.swift
jeppes: `Response` is a subset of each of things you mentioned. I don't think it makes sense to use the same name.

MobiusCore/Source/EffectHandlers/EffectExecutor.swift
```diff
@@ -0,0 +1,100 @@
...
+
+    // Keep track of each received effect's state.
+    // When an effect has completed, it should be removed from this dictionary.
+    // When disposing this effect handler, all entries must be removed.
+    private var handlingEffects: [Int64: EffectHandlingState<Event>] = [:]
+    private var nextID = Int64(0)
+
+    init(handleInput: @escaping (Effect, Response<Event>) -> Disposable) {
+        self.handleEffect = handleInput
+    }
```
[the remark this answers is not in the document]
jeppes: Any `Response`s that were still inflight would be ended, which would do nothing since they only hold a weak reference to `EffectExecutor` objects, and all the work they do in `end` is remove the references held by `EffectExecutor`.

Do you think we should be disposing in `deinit`?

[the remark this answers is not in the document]
JensAyton: As a general rule, I think disposables should either dispose or assert that they’re already disposed in deinit.

MobiusCore/Source/EffectHandlers/Response.swift
```diff
@@ -0,0 +1,70 @@
...
+
+    public init(
+        onSend: @escaping (T) -> Void,
+        onEnd: @escaping () -> Void
+    ) {
+        self.onSend = onSend
+        self.onEnd = onEnd
+    }
+
+    public func end() {
```
kmcbride: Would it be problematic for this function to take an optional output?
```swift
public func end(_ finalOutput: T? = nil) {
    …
    if runOnEnd {
        if let output = finalOutput {
            onSend(output)
        }
        onEnd()
    }
}
```
Perhaps the idea was already discussed, but it would make the API usage a little cleaner.

MobiusCore/Source/EffectHandlers/Response.swift
pettermahlen: Not "a subset of each", but a subset of all, I guess you mean? I still think it makes sense to share the name, mainly because they fill exactly the same niche - enabling reporting back about the progress of handling an effect.

MobiusCore/Source/EffectHandlers/Response.swift
pettermahlen: Cleaner for the specific case when you want to send a single event, I guess you mean? The drawback, ofc, is that it gives the `end` method two responsibilities as opposed to just one. Also, it creates a difference between this API and the (current) Kotlin proposal, and prior art in the form of the design from reactive-streams.org. The advantage is that it makes one use case a one-liner instead of a two-liner. I don't think that's a huge advantage, but it's definitely an advantage.

kmcbride: Correct; "cleaner" in the sense that the response could always be ended in one line, regardless of how many events it sent overall.

In any case, it was based on an initial reaction to the usage examples, so if it breaks API parity and/or clashes with the design principles, those bothered enough by making two function calls could just as easily extend the Response class to wrap them in a single call.

MobiusCore/Source/EffectHandlers/Response.swift
jeppes: I mean "of each".

`Effect.Callback` would not work since we don't have an `Effect` type. `EffectHandler.Callback` would work, but due to how Swift works we would also need to expose the underlying `Response` type, or whatever we choose to call it.

`Subscriber` would work too, but I don't think it's a very helpful name in this context.

jeppes: `EffectResponse` and `EffectCallback` would work. Would these be preferable?

pettermahlen: @anawara opinions?

marshhxx: I agree we should use the same names. I don't really have strong opinions in which name to pick although `Callback` feels a bit more natural to me. Showing some code below to get a feeling on how it may look.

With `EffectResponse` it will look like:
```
interface Effect<T> {
    interface Response<T> {
        fun start(disposable: Disposable)
        fun send(value: T)
        fun end()
    }
    fun execute(response: Response<T>)
}
```
and when used:
```
fun <T> fromValue(value: T): Effect<T> = object : Effect<T> {
    override fun execute(response: Response<T>) = with(response) {
        start(Disposable {})
        send(value)
        end()
    }
}
```
vs `EffectCallback`
```
interface Effect<T> {
    interface Callback<T> {
        fun onStart(disposable: Disposable)
        fun onValue(value: T)
        fun onComplete()
    }
    fun execute(callback: Callback<T>)
}
```
and when used:
```
fun <T> fromValue(value: T): Effect<T> = object : Effect<T> {
    override fun execute(callback: Callback<T>) = with(callback) {
        onStart(Disposable {})
        onValue(value)
        onComplete()
    }
}
```

jeppes: I prefer `Response` because it tells you what it does as opposed to what strategy was used to implement it.

pettermahlen: A disadvantage of `Response` is it is usually used synchronously/for a single response, https://en.wikipedia.org/wiki/Request%E2%80%93response. This is an asynchronous use case, with multiple values possibly associated with an originating request/effect.

anawara: I agree with Petter, the name Response indicates that it's an output that I can inspect to see a result. It indicates that the operation has already been completed and this is the wrapper type holding it. However, what we're passing in here is a continuation/callback. It's also a little odd that I pass in a response to the execute/handle function. It reads really awkwardly

jeppes: That's fair

jeppes: I'll change it to `EffectCallback`

MobiusCore/Source/EffectHandlers/EffectRouterDSL.swift
```diff
@@ -28,47 +28,38 @@ public extension EffectRouter where Input: Equatable {
 }
 
 public extension PartialEffectRouter {
+    /// Route to the an anonymous  `EffectHandler` defined by the `handle` closure.
```
pettermahlen:
```suggestion
    /// Route to the anonymous `EffectHandler` defined by the `handle` closure.
```

MobiusCore/Source/EffectHandlers/ThreadSafeConnectable.swift
```diff
@@ -0,0 +1,79 @@
...
+    private var output: Consumer<Event>?
+    private var connection: Connection<Effect>?
+
+    init<Conn: Connectable>(
+        connectable: Conn
+    ) where Conn.InputType == Effect, Conn.OutputType == Event {
+        self.connectable = AnyConnectable(connectable)
+    }
+
+    var isConnected = false
```
pettermahlen: This var doesn't seem to be used anywhere, it should be deleted.

MobiusCore/Source/EffectHandlers/ThreadSafeConnectable.swift
```diff
@@ -0,0 +1,79 @@
...
+
+            return Connection(
+                acceptClosure: accept,
+                disposeClosure: dispose
+            )
+        }
+    }
+
+    private func accept(_ effect: Effect) {
+        if let connection = lock.synchronized(closure: { connection }) {
```
pettermahlen: Wouldn't it be more readable to do the same here as is done in `dispatch`?

MobiusCore/Test/EffectHandlers/EffectHandlerTests.swift
```diff
@@ -38,24 +38,23 @@ class EffectHandlerTests: QuickSpec {
...
 
             beforeEach {
-                effectHandler = EffectHandler(
-                    handle: handleEffect,
-                    disposable: AnonymousDisposable {}
-                )
+                effectHandler = AnyEffectHandler(handle: handleEffect)
                 receivedEvents = []
-                let output = { (event: Event) in receivedEvents.append(event) }
+                let response = EffectCallback<Event>(
```
pettermahlen:
```suggestion
                let callback = EffectCallback<Event>(
```

MobiusCore/Test/EffectHandlers/EffectHandlerTests.swift
```diff
@@ -38,24 +38,23 @@ class EffectHandlerTests: QuickSpec {
...
+                    onSend: { event in
+                        receivedEvents.append(event)
+                    },
+                    onEnd: {}
+                )
                 executeEffect = { effect in
-                    effectHandler.handle(effect, output)
+                    _ = effectHandler.handle(effect, response)
```
pettermahlen:
```suggestion
                    _ = effectHandler.handle(effect, callback)
```

MobiusCore/Test/EffectHandlers/EffectHandlerTests.swift
```diff
@@ -70,39 +69,26 @@ class EffectHandlerTests: QuickSpec {
...
-            }
-
-            it("disposing is idempotent") {
-                expect(disposed).to(beFalse())
-
-                effectHandler.disposable.dispose()
-                effectHandler.disposable.dispose()
-                effectHandler.disposable.dispose()
+                }
+                let response = EffectCallback<Event>(onSend: { _ in }, onEnd: {})
```
pettermahlen:
```suggestion
                let callback = EffectCallback<Event>(onSend: { _ in }, onEnd: {})
```

MobiusCore/Test/EffectHandlers/EffectHandlerTests.swift
```diff
@@ -70,39 +69,26 @@ class EffectHandlerTests: QuickSpec {
...
+                effectHandler.handle(.effect1, response).dispose()
```
pettermahlen:
```suggestion
                effectHandler.handle(.effect1, callback).dispose()
```

MobiusCore/Test/EffectHandlers/EffectRouterDSLTests.swift
```diff
@@ -36,29 +36,92 @@ private enum Event: Equatable {
...
+            }
+
+            it("should not be disposed if `end()` was called") {
+                connection.accept(1)
+                connection.dispose()
+
+                expect(wasDisposed).to(beFalse())
+            }
+
+            it("should be disposed if the connection is disposed before `end()` is called") {
```
pettermahlen: The description of this test case doesn't match what's being tested. I think the description is right, not sure why `wasDisposed` is false?

MobiusCore/Test/EffectHandlers/EffectRouterDSLTests.swift
```diff
@@ -36,29 +36,92 @@ private enum Event: Equatable {
...
+        context("An EffectHandler which calls end after some time") {
+            var wasDisposed: Bool!
+            var connection: Connection<Int>!
+            var end: (() -> Void)!
+
+            beforeEach {
+                wasDisposed = false
+                end = { fail("End should have been set") }
+                connection = EffectRouter<Int, ()>()
+                    .routeEffects(equalTo: 1).to { _, response in
```
pettermahlen: I guess `response` should be changed to `callback` everywhere in this file (or pretty much).

MobiusCore/Test/EffectHandlers/EffectRouterDSLTests.swift
```diff
@@ -100,43 +162,26 @@ class EffectRouterDSLTests: QuickSpec {
...
-                        wasDisposed = true
-                    }
-                )
+
                 let payload: (Effect) -> Effect? = { $0 == .effect1 ? .effect1 : nil }
                 let dslHandler = EffectRouter<Effect, Event>()
-                    .routeEffects(withPayload: payload).to(effectHandler)
+                    .routeEffects(withPayload: payload).to { effect, response in
+                        expect(effect).to(equal(.effect1))
+                        response.send(.eventForEffect1)
```
pettermahlen: Should this call `end`? Probably doesn't matter for the purposes of this PR. 

I am starting to think that maybe there should in fact be a way to send a single event, because it is easier. It's breaking the single responsibility principle, but following the principle "don't make the client do anything the library could" (https://dl.acm.org/doi/10.1145/1176617.1176622). But that is something we can address later, no need to do it in this PR.

MobiusCore/Test/EffectHandlers/EffectRouterTests.swift
```diff
@@ -49,22 +49,18 @@ class EffectRouterTests: QuickSpec {
                 receivedEvents = []
                 disposed1 = false
                 disposed2 = false
-                let effectHandler1 = EffectHandler<Effect, Event>(
-                    handle: { _, dispatch in
-                        dispatch(.eventForEffect1)
-                    },
-                    disposable: AnonymousDisposable {
+                let effectHandler1 = AnyEffectHandler<Effect, Event> { _, response in
```
pettermahlen: Again s/response/callback/g.

MobiusCore/Test/EffectHandlers/ResponseTests.swift
```diff
@@ -0,0 +1,92 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import MobiusCore
+import Nimble
+import Quick
+
+class ResponseTests: QuickSpec {
```
pettermahlen: Rename `Response`/`response` to `Callback`/`callback`

MobiusCore/Test/EventRouterDisposalLogicalRaceRegressionTest.swift
```diff
@@ -113,20 +113,17 @@ private class EffectCollaborator {
...
-            handle: { _, dispatch in
-                cancellationToken?.cancel()
-                cancellationToken = self.asyncDoStuff {
-                    dispatch(replyEvent)
-                }
-            },
-            disposable: AnonymousDisposable {
-                cancellationToken?.cancel()
+    func makeEffectHandler<Effect, Event>(replyEvent: Event) -> AnyEffectHandler<Effect, Event> {
+        return AnyEffectHandler<Effect, Event> { _, response in
```
pettermahlen: `response` -> `callback`

MobiusCore/Test/NonReentrancyTests.swift
```diff
@@ -113,14 +113,15 @@ class NonReentrancyTests: QuickSpec {
 
             context("when effect handler posts events through consumer") {
                 beforeEach {
-                    handleEffect = { effect, eventConsumer in
+                    handleEffect = { effect, response in
```
pettermahlen:
```suggestion
                    handleEffect = { effect, callback in
```

MobiusCore/Test/NonReentrancyTests.swift
```diff
@@ -113,14 +113,15 @@ class NonReentrancyTests: QuickSpec {
...
                         log("handle enter - effect: \(effect)")
                         defer {
                             log("handle exit - effect: \(effect)")
                         }
 
-                        eventConsumer(.increment)
-                        eventConsumer(.increment)
+                        response.send(.increment)
```
pettermahlen:
```suggestion
                        callback.send(.increment)
```

pettermahlen: and below.

MobiusCore/Test/EffectHandlers/EffectRouterDSLTests.swift
jeppes: This is testing that the `EffectRouter` is disposed. It will not be disposed it it was ended (which is tested separately).It is confusing though, because this is also testing that it can send events.

I will split this into two test cases to address this confusion

jeppes: `toEvent` is what you would use for sending a single event :)

pettermahlen: Re `toEvent`: yes, if you know up front that it'll always be a single event. Maybe there are conditions where it'll be one, none, or many. Again, I think we should leave that for later. The existence of `toEvent` makes it even less urgent to address, and when we have some usage of this version, we'll know more about whether we'll want a special-cased solution for single event/end in the callback as well.

jeppes: Ok, I understand. Sounds reasonable. Adding a version of `end` which takes an event would be a compatible change - so, like you said, let's revisit this

MobiusCore/Source/EffectHandlers/EffectHandler.swift
```diff
@@ -17,16 +17,47 @@
...
+///
+/// Note: `EffectHandler` should be used in conjunction with an `EffectRouter`.
+public protocol EffectHandler {
+    associatedtype Effect
+    associatedtype Event
+
+    /// Handle an `Effect`.
+    /// To output events, call `callback.send`.
+    /// When you are done handling `input`, be sure to call `callback.end()` to prevent memory leaks.
+    /// If it does not make sense to finish handling an effect, you should be using a `Connectable` instead of this protocol.
```
JensAyton: What kind of cases do you see this applying to?

JensAyton:
Sorry for coming along late, but I really dislike the name “callback” for a thing with two callbacks. :-/

MobiusCore/Source/EffectHandlers/ThreadSafeConnectable.swift
```diff
@@ -0,0 +1,78 @@
...
+            shouldDispose = output != nil
+            output = nil
+        }
+        if shouldDispose {
+            connection?.dispose()
+            lock.synchronized {
+                connection = nil
+            }
+        }
+    }
```
JensAyton: Simpler and less potentially racy:
```swift
private func dispose() {
    var disposeConnection: (() -> Void)?
    lock.synchronized {
        output = nil
        disposeConnection = connection?.dispose
        connection = nil
    }
    disposeConnection?()
}
```

MobiusCore/Source/EffectHandlers/EffectHandler.swift
jeppes: @pettermahlen and I had a discussion about this, so maybe he can provide some insight too.

The way I see it, `Connectable` is still useful in the cases where a clear connection between effects and events cannot be made. For example, an effect handler which is used as an adapter to talk between two features. Such an effect handler would translate effects from one loop into messages to the other feature, and relay messages from the other feature back into the loop.

MobiusCore/Source/EffectHandlers/EffectExecutor.swift
```diff
@@ -0,0 +1,111 @@
...
+            // Any events produced as a result of handling the effect will be sent to this class's `output` consumer,
+            // unless it has already been disposed.
+            onSend: { [weak self] event in self?.output?(event) },
+            // Once an effect has been handled, remove the reference to its callback and disposable.
+            onEnd: { [weak self] in self?.delete(id: id) }
+        )
+
+        let disposable = handleEffect(effect, callback)
+
+        store(id: id, callback: callback, disposable: disposable)
```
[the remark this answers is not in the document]
jeppes: I don't think it's worth the effort for this

jeppes:
@JensAyton 
> Sorry for coming along late, but I really dislike the name “callback” for a thing with two callbacks. :-/

That's fair. Any other ideas?

MobiusCore/Source/EffectHandlers/EffectCallback.swift
```diff
@@ -55,16 +54,14 @@ public final class EffectCallback<Output> {
     }
 
     public func send(_ output: Output) {
-        lock.synchronized {
```
jeppes: We cannot know how long this lock will be held. Calls to the output consumer are serialized by the framework anyway. The only thing we need to ensure here is that `end` is not called while the output consumer is being used.

JensAyton:
I don’t have a satisfactory naming suggestion right now. There’s time for bikeshedding later.

PR #104, opened 2020-01-21
Add a single API for unwrapping Enums and routing to Enum constants

jeppes:
This PR introduces a single way of routing to any `enum`-based effect.
This is what it would look like:
```swift
enum Effect: Equatable {
   case a
   case b(string: String)
   case c(left: String, right: String)
}
​
let router = EffectRouter<Effect, Event>()
    .routeCase(Effect.a).to { a in ... }
    .routeCase(Effect.b).to { string in ... }
    .routeCase(Effect.c).to { (left, right) in ... }
```
Reflection is used to unwrap effects with associated data, so no more unwrapping boilerplate or code generation is required. Thank you @mbrandonw and @stephencelis for your work on this code!

One issue is that `enum`s which do not have associated data must be equatable to work with this. The error for this luckily is very intuitive.

@pettermahlen @JensAyton


stephencelis:
Definitely nice to be able to extract associated values with case path reflection! I think my only question around this kind of routing is whether or not it's important to exhaustively switch over effects so that the compiler can catch missing and redundant handlers rather than pushing that check to runtime.

jeppes:
Both cases are possible.

PR #105, opened 2020-01-21
Fix logical race caused by asynchronous view connection cleanup

JensAyton:
View connections are disposed asynchronously on the view queue (typically the main queue). If an event is dispatched on the view queue after stopping the Mobius controller, we would hit an assertion. Unlike previous logic races, this could reasonably be worked around in client code, but it would introduce annoying coupling and synchronization work which we can handle generically using our existing dropping-events-on-the-floor technique.

@jeppes

### DIFF
--- a/MobiusCore/Source/ConnectableConvenienceClasses/AsyncDispatchQueueConnectable.swift
+++ b/MobiusCore/Source/ConnectableConvenienceClasses/AsyncDispatchQueueConnectable.swift
@@ -36,7 +36,6 @@ final class AsyncDispatchQueueConnectable<InputType, OutputType>: Connectable {
         case pendingDispose
         case disposed
     }
-    private var disposalStatus = Synchronized(value: DisposalStatus.notDisposed)
 
     init(
         _ underlyingConnectable: AnyConnectable<InputType, OutputType>,
@@ -54,7 +53,9 @@ final class AsyncDispatchQueueConnectable<InputType, OutputType>: Connectable {
     }
 
     func connect(_ consumer: @escaping (OutputType) -> Void) -> Connection<InputType> {
-        let connection = underlyingConnectable.connect { [disposalStatus] value in
+        let disposalStatus = Synchronized(value: DisposalStatus.notDisposed)
+
+        let connection = underlyingConnectable.connect { value in
             // Don’t forward if we’re currently waiting to dispose the connection.
             //
             // NOTE: the underlying consumer must be called inside the critical region accessing disposalStatus, or we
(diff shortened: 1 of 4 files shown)


MobiusCore/Test/MobiusControllerTests.swift
```diff
@@ -142,6 +142,17 @@ class MobiusControllerTests: QuickSpec {
                             controller.start()
                             expect(controller.running).to(beTrue())
                         }
+                        fit("should ignore events sent while view disposal is pending") {
```
rastersize: This should probably be `it` and not `fit` for master 🙂

MobiusCore/Source/ConnectableConvenienceClasses/AsyncDispatchQueueConnectable.swift
```diff
@@ -44,17 +54,33 @@ final class AsyncDispatchQueueConnectable<InputType, OutputType>: Connectable {
     }
 
     func connect(_ consumer: @escaping (OutputType) -> Void) -> Connection<InputType> {
-        let connection = underlyingConnectable.connect(consumer)
+        let connection = underlyingConnectable.connect { [disposalStatus] value in
+            // Don’t forward if we’re currently waiting to dispose the connection.
+            //
+            // NOTE: the underlying consumer must be called inside the criticial region accessing disposalStatus, or we
```
pettermahlen:
```suggestion
            // NOTE: the underlying consumer must be called inside the critical region accessing disposalStatus, or we
```

MobiusCore/Source/ConnectableConvenienceClasses/AsyncDispatchQueueConnectable.swift
```diff
@@ -44,17 +54,33 @@ final class AsyncDispatchQueueConnectable<InputType, OutputType>: Connectable {
...
 
         return Connection(
             acceptClosure: { [acceptQueue] input in
                 acceptQueue.async {
                     connection.accept(input)
                 }
             },
-            disposeClosure: { [acceptQueue] in
+            disposeClosure: { [acceptQueue, disposalStatus] in
+                disposalStatus.value = .pendingDispose
```
pettermahlen: What happens if there are multiple calls to `dispose`?

MobiusCore/Test/MobiusControllerTests.swift
JensAyton: I was just thinking yesterday that we should lint for this…

MobiusCore/Source/ConnectableConvenienceClasses/AsyncDispatchQueueConnectable.swift
JensAyton: Updated to assert on that… and scope the `disposalStatus` to the individual connection, not the connectable. Oops!

MobiusCore/Test/MobiusControllerTests.swift
rastersize: @JensAyton We lint for it in S4A using [Danger](https://danger.systems) and the following snippet:

```ruby
# Make sure we aren't focusing a specific test and ignoring all others. We don't want to let
# testing shortcuts get into master.
(git.modified_files + git.added_files).each do |file|
  next unless File.file?(file)
  if file.downcase.include?('/tests/') && file.end_with?('.swift')
    contents = File.read(file)
    fail("`xit` or `fit` left in tests (#{file})") if contents =~ /^\w*[xf]it/
    fail("`fdescribe` left in tests (#{file})") if contents =~ /^\w*fdescribe/
    fail("`fitBehavesLike` left in tests (#{file})") if contents =~ /^\w*fitBehavesLike/
  end
end
```

PR #106, opened 2020-01-22
Document all functions in `EffectCallback`

jeppes:
This should help clarify the semantics around the order in which `send` and `end` can be called.

@pettermahlen @JensAyton

### DIFF
--- a/MobiusCore/Source/EffectHandlers/EffectCallback.swift
+++ b/MobiusCore/Source/EffectHandlers/EffectCallback.swift
@@ -24,7 +24,7 @@ import Foundation
 /// Sending output is done with `.send` and signaling completion is done with `.end`. You can also end in conjunction
 /// with sending output using `.end(with:)`.
 ///
-/// Note: Once `.end` has been called (from any thread), all operations on this object will be no-ops.
+/// Note: Once either `.end` or  `.end(with:)`  have been called (from any thread), all operations on this object will be no-ops.
 /// Note: The closure provided in `onEnd` will only be called once when `.end` is called on this object.
 public final class EffectCallback<Output> {
     private let onSend: (Output) -> Void
@@ -52,7 +52,7 @@ public final class EffectCallback<Output> {
         self.onEnd = onEnd
     }
 
-    /// Invalidate this callback, and deinit the associated `EffectHandler`.
+    /// Invalidate this callback.
     /// Note: any calls to `end`, `end(with:)` or `send(_:)` will be no-ops after this function has been called.
     public func end() {
         var runOnEnd = false
@@ -66,24 +66,31 @@ public final class EffectCallback<Output> {
     }
 
     /// Send a number of events to the Mobius loop, then `end()` this callback.
-    /// Note: Calling this function after calling `.end()` is a no-op.
+    /// Note: After calling this function, all operations on this object will be no-ops.
     /// - Parameter outputs: The events which should be sent to the loop.
     public func end(with outputs: Output...) {
         end(with: outputs)
     }
 
-    /// Send a number of events to the Mobius loop, then `end()` this callback.
-    /// Note: Calling this function after calling `.end()` is a no-op.
+    /// Send a number of events to the Mobius loop and `end()` this callback.
+    /// Note: After calling this function, all operations on this object will be no-ops.
     /// - Parameter outputs: The events which should be sent to the loop.
     public func end(with outputs: [Output]) {
-        for output in outputs {
-            send(output)
+        var runOnEnd = false
+        _ended.mutate {
+            runOnEnd = !$0
+            $0 = true
         }
-        end()
+        for output in outputs {
+            onSend(output)
+        }
+        if runOnEnd {
+            onEnd()
+        }
     }
 
     /// Send an event to the Mobius loop.
-    /// Note: Calling this function after calling `.end()` is a no-op.
+    /// Note: Calling this function after calling `.end()` or `.end(with:)` is a no-op.
     /// - Parameter output: the event that should be sent to the loop.
     public func send(_ output: Output) {
         _ended.mutate {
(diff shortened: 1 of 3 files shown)


MobiusCore/Source/EffectHandlers/EffectHandler.swift
```diff
@@ -28,13 +28,16 @@ public protocol EffectHandler {
 
     /// Handle an `Effect`.
     /// To output events, call `callback.send`.
-    /// When you are done handling `input`, be sure to call `callback.end()` to prevent memory leaks.
-    /// If it does not make sense to finish handling an effect, you should be using a `Connectable` instead of this protocol.
+    /// Call `callback.end()` once the effect has been handled to prevent memory leaks.
     ///
+    /// This returns a `Disposable` which tears down any resources that is being used by this effect handler. This `Disposable` will
```
pettermahlen:
```suggestion
    /// This returns a `Disposable` which tears down any resources used to handle this effect. This `Disposable` will
```

pettermahlen: And maybe it's not even about tearing down resources, but cancelling the handling of this effect?

MobiusCore/Source/EffectHandlers/EffectHandler.swift
```diff
@@ -28,13 +28,16 @@ public protocol EffectHandler {
...
-    /// When you are done handling `input`, be sure to call `callback.end()` to prevent memory leaks.
-    /// If it does not make sense to finish handling an effect, you should be using a `Connectable` instead of this protocol.
+    /// Call `callback.end()` once the effect has been handled to prevent memory leaks.
     ///
+    /// This returns a `Disposable` which tears down any resources used to handle this effect. This `Disposable` will
+    /// not be called if `callback.end()` has already been called.
+    ///
+    /// Note: If it does not make sense to finish handling an effect, you should be using a `Connectable` instead of this protocol.
     /// Note: When being disposed by Mobius, the `Disposable` you return will be called before Mobius calls `callback.end()`.
     /// Note: Mobius will not dispose the returned `Disposable` if `callback.end()` has already been called.
```
JensAyton: Duplicated info?

MobiusCore/Source/EffectHandlers/EffectCallback.swift
```diff
@@ -23,18 +23,26 @@ import Foundation
...
-/// Note: The closure you provide in `onEnd` will only be called once when `.end` is called on this object.
+/// Note: Once `.end` has been called (from any thread), all operations on this object will be no-ops.
+/// Note: The closure provided in `onEnd` will only be called once when `.end` is called on this object.
 public final class EffectCallback<Output> {
     private let onSend: (Output) -> Void
     private let onEnd: () -> Void
 
+    /// Determine if this callback has been ended.
+    /// This can be called safely from any thread.
+    /// Note: Once this variable is `false`, it will never be `true` again.
```
pettermahlen:
```suggestion
    /// Note: Once this variable is `true`, it will never be `false` again.
```

MobiusCore/Source/EffectHandlers/EffectCallback.swift
```diff
@@ -43,6 +51,8 @@ public final class EffectCallback<Output> {
         self.onEnd = onEnd
     }
 
+    /// Invalidate this callback, and deinit the associated `EffectHandler`.
```
pettermahlen: Can that be known? As in, I guess deinit will happen when the memory is freed?

MobiusCore/Source/EffectHandlers/EffectCallback.swift
```diff
@@ -54,17 +64,26 @@ public final class EffectCallback<Output> {
...
+    /// - Parameter outputs: The events which should be sent to the loop.
     public func end(with outputs: Output...) {
         end(with: outputs)
     }
 
+    /// Send a number of events to the Mobius loop, then `end()` this callback.
+    /// Note: Calling this function after calling `.end()` is a no-op.
+    /// - Parameter outputs: The events which should be sent to the loop.
     public func end(with outputs: [Output]) {
         for output in outputs {
```
pettermahlen: This actually means that someone else might call `end()` and interrupt the flow of outputs, or that this set of outputs might be mixed with another invocation of `send`. I think it would be better to set `ended` to true at the beginning of this method, then invoke the `onSend` closure instead of the `send` method.

MobiusCore/Source/EffectHandlers/EffectCallback.swift
```diff
@@ -23,18 +23,26 @@ import Foundation
 /// Sending output is done with `.send` and signaling completion is done with `.end`.
 /// You can also end in conjunction with sending output using `.end(with:)`
 ///
-/// Note: Once `.end` has been called (from any thread), the closure you provide in `onSend` will no longer be called.
-/// Note: The closure you provide in `onEnd` will only be called once when `.end` is called on this object.
+/// Note: Once `.end` has been called (from any thread), all operations on this object will be no-ops.
```
pettermahlen: For clarity, maybe this should say something like `end` or `end(with:)` ?

MobiusCore/Source/EffectHandlers/EffectCallback.swift
```diff
@@ -54,17 +64,33 @@ public final class EffectCallback<Output> {
...
+    /// Send a number of events to the Mobius loop and `end()` this callback.
+    /// Note: After calling this function, all operations on this object will be no-ops.
+    /// - Parameter outputs: The events which should be sent to the loop.
     public func end(with outputs: [Output]) {
+        var runOnEnd = false
+        _ended.mutate {
+            runOnEnd = !$0
+            $0 = true
+        }
         for output in outputs {
```
pettermahlen: I think this needs to be inside the `if runOnEnd` condition? Would be good with a test that does two `end(with:)` calls, it looks to me like that would today lead to both sets being sent.

jeppes: Good catch

JensAyton:
Oops.

jeppes:
@pettermahlen, thanks for your review, sorry I needed to rebase to resolve the conflicts.

PR #108, opened 2020-01-23
Misc documentation and cleanup

JensAyton:
* Documents some simple stuff
* Consistently wraps touched documentation at 120 columns
* Deprecates some low-value typedefs and methods
  - I don’t feel that `next. hasEffects` is significantly more expressive than `!next.effects.isEmpty`. If `isEmpty` didn’t exist and you had to use `count != 0` it would perhaps carry its weight.
  - I don’t see any reason for `BrokenConnection.accept` and `.dispose` to be accessed directly.
  - `ConnectClosure` was used in one place in API. `Connection.ValueType` was identical to `Connection.Value` and not used at all.
  - Names like `ValueType` and `InputType` are non-idiomatic.

@jeppes

### DIFF
--- a/MobiusCore/Source/BrokenConnection.swift
+++ b/MobiusCore/Source/BrokenConnection.swift
@@ -19,16 +19,24 @@
 
 import Foundation
 
+/// A helper to construct `Connection`s that will fail when used.
+///
+/// `BrokenConnection.connection()` is used when functions return `Connection` fail an assertion. The resulting
+/// connection will trigger an assertion whenever its `accept` or `dispose` methods are called.
 public enum BrokenConnection<ValueType> {
-    public static func accept(_ value: ValueType) {
+    static func _accept(_ value: ValueType) {
         MobiusHooks.onError("'accept' called on invalid connection")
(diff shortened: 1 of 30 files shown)


MobiusCore/Source/BrokenConnection.swift
```diff
@@ -19,16 +19,24 @@
...
 import Foundation
 
-public enum BrokenConnection<ValueType> {
-    public static func accept(_ value: ValueType) {
+/// A helper to construct `Connection`s that will fail when used.
+///
+/// `BrokenConnection.connection()` is used when functions return `Connection` fail an assertion. The resulting
+/// connection will trigger an assertion whenever its `accept` or `dispose` methods are called.
+public enum BrokenConnection<Value> {
+    static func _accept(_ value: Value) {
```
jeppes: I believe these can actually be private

JensAyton: No, because they’re called from BackwardsCompatiblity.swift. Should be in future though. :-)

PR #111, opened 2020-01-25
Adopt renamed: parameter on deprecation attributes

JensAyton:
Apparently this is a thing. (The benefit is that deprecation warnings get fixits.)

### DIFF
--- a/MobiusCore/Source/BackwardsCompatibility.swift
+++ b/MobiusCore/Source/BackwardsCompatibility.swift
@@ -19,7 +19,7 @@
 
 import Foundation
 
-@available(*, deprecated, message: "use Initiate instead")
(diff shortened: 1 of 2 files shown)


MobiusCore/Source/BackwardsCompatibility.swift
```diff
@@ -90,12 +90,12 @@ public extension MobiusController {
         )
     }
 
-    @available(*, deprecated, message: "use model instead")
+    @available(*, deprecated, renamed: "model")
```
JensAyton: In this case, the fixit will change `loop.getModel()` to `loop.model()`, which is wrong, but the resulting error is reasonably clear.

jeppes:
Very cool!

PR #114, opened 2020-02-14
Idea: Allow EffectRouter to be passed to Mobius.loop directly

JensAyton:
The `asConnectable` at the end of an `EffectRouter` setup is an annoying abstraction leak. Supporting an array of multiple `update` types _and_ multiple `effectHandler` types is pretty unattractive, though.

Here’s a third option: use an underscored protocol to provide a common supertype to `EffectRouter<Effect, Event>` and `Connectable where Input == Effect, Output == Event`. I’m not convinced myself, the extra associated types are a pain, but I thought it was worth looking at.

@jeppes

### DIFF
--- a/MobiusCore/Source/EffectHandlers/EffectRouter.swift
+++ b/MobiusCore/Source/EffectHandlers/EffectRouter.swift
@@ -72,6 +72,12 @@ public struct EffectRouter<Effect, Event> {
     }
 }
 
+extension EffectRouter: _EffectHandlerConvertible {
+    public func _asEffectHandlerConnectable() -> AnyConnectable<Effect, Event> {
+        return compose(routes: routes)
+    }
+}
+
 public struct _PartialEffectRouter<Effect, Payload, Event> {
     fileprivate let routes: [Route<Effect, Event>]
     fileprivate let path: (Effect) -> Payload?
--- a/MobiusCore/Source/Connectable.swift
+++ b/MobiusCore/Source/Connectable.swift
@@ -26,7 +26,7 @@ import Foundation
 ///
 /// Alternatively used in `MobiusController.connectView(_:)` to connect a view to the controller. In that case, the
 /// incoming values will be models, and the outgoing values will be events.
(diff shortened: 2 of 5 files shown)


JensAyton:
This also allows effect routers to be written without explicitly naming the effect and event types, although I admit it ends up looking a bit weird:
```swift
loop = Mobius.loop(
    update: update,
    effectHandler: EffectRouter()
        .routeEffects(equalTo: .testEffect).to(testEffectHandler)
)
    .start(from: 0)
```

rastersize:
I very much like this idea! The `asConnectable` thing has bothered me ever so slightly 😸

MobiusCore/Source/EffectHandlers/EffectRouter.swift
```diff
@@ -72,6 +72,12 @@ public struct EffectRouter<Effect, Event> {
     }
 }
 
+extension EffectRouter: _EffectHandlerConvertible {
+    public func _asEffectHandlerConnectable() -> AnyConnectable<Effect, Event> {
```
jeppes: Should the `_asEffectHandlerConnectable()` methods be public?

JensAyton: Yes; this is necessary since protocols can’t have multiple levels of access control. In order to publicly conform to the protocol the method must be implemented publicly.

JensAyton:
@jeppes Do you expect we’ll need the `Connectable` version in the long term, or is it transitional?

jeppes:
@JensAyton I think we either need to keep `Connectable` around or introduce exit-effects to the loop

kmcbride:
Sorry to bump an old thread, but was something like this ruled out?
```swift
extension EffectRouter: Connectable {
    public func connect(_ consumer: @escaping Consumer<Event>) -> Connection<Effect> {
        return compose(routes: routes).connect(consumer)
    }
}
```
It seems like it would "just work" and avoid the need for any other API changes.

PR #116, opened 2020-02-17
Allow EffectRouter to be passed to Mobius.loop directly

JensAyton:
Alternative to #114 .

The downside of this approach is that there are now four overloads of `Mobius.loop`, which could get nasty if we want to add more dimensions of variants, or just change the interface (for instance, using defaulted arguments instead of the three remaining `with` functions).

I prefer this approach if we view the explicit `Connectable` approach as transitional. If we want to keep all the variants forever, I lean towards #114 .

@jeppes @rastersize


jeppes:
Hmm, I'm not sure about this. I think there are good arguments for both.
Any thoughts, @pettermahlen?

PR #118, opened 2020-02-17
Make MobiusController view connection optional

JensAyton:
Since `MobiusController` now represents “starty-stoppy MobiusLoop running on a background queue” rather than “ViewController glue for MobiusLoop”, it no longer makes sense to require `connectView`.

@jeppes


JensAyton:
Apparently the linter wants me to remove some comments from MobiusController. 🙄

JensAyton:
As far as I can understand, the coverage warning is complaining about lacking coverage on a closing brace and some code that’s `#if`ed out.

In any case, coverage goes back to green in #121, which is a continuation of this branch.

PR #126, opened 2020-03-09
Update CI

dflems:
Split up SPM build/test from Xcode build/test. Let's see if the issues from #59 have resolved themselves.

> **UPDATE:** They didn't, but I fixed it by patching carthage deps to build with the legacy build system. Sort of weird but it works.


Tools/helpers.sh
```diff
@@ -31,9 +31,23 @@ dump_log() {
   echo ""
 }
 
+patch_to_legacy_build_system() {
+  /usr/libexec/PlistBuddy -c "Delete :BuildSystemType" "$1" 2>/dev/null
```
BalestraPatrick: 😅

PR #127, opened 2020-03-10
GH Actions: Add caching

dflems:
This should make things build a little faster in CI

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,7 +39,6 @@ playground.xcworkspace
 # Add this line if you want to avoid checking in source code from Swift Package Manager dependencies.
 # Packages/
 # Package.pins
-Package.resolved
 .build/
 .swiftpm/
 
--- a/.github/workflows/pull.yaml
+++ b/.github/workflows/pull.yaml
@@ -7,6 +7,12 @@ jobs:
     runs-on: macOS-latest
     steps:
     - uses: actions/checkout@v1
+    - uses: actions/cache@v1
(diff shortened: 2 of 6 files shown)


dflems:
Before:
image

After:
image (1)

.gitignore
```diff
@@ -39,7 +39,6 @@ playground.xcworkspace
 # Add this line if you want to avoid checking in source code from Swift Package Manager dependencies.
 # Packages/
 # Package.pins
-Package.resolved
```
JensAyton: I’m not clear on the workflow consequences of this change.

dflems: Not sure why it would matter?

JensAyton: My expectation would be that before this change, people would get the latest compatible-according-to-semver versions of dependencies, but with them, we’re pinned to specific revisions. That implies that someone™ should periodically be updating the pinned revisions.

dflems: I assumed that would only be if you're building _this_ project, not when using mobius as a dependency of your own project. AFAIK it only matters for the top-level package. I could verify that if you have a concern there.

JensAyton: That pretty much has to be true, but I don’t particularly like growing obsolescence in CI (and my local builds in Xcode 11.4) either

dflems: I get that. I think it's good to pin _our_ dependencies in this repo to avoid weird discrepancies (builds fine on your machine but not mine).

PR #131, opened 2020-03-18
Automatic MobiusController lifecycle management in `UIViewController`s

jeppes:
This PR makes it possible to bind the lifecycle of a `MobiusController` to a `UIViewController`.

The controller is started by calling the `useController` method (shown below), and can be stopped manually at any point. If it is not manually stopped, it will stop when the view controller is deinitialized.

Example:

```swift
class ViewController: UIViewController {
    let controller = Mobius.loop(...).makeController(from: ...)
    let viewBinder: AnyConnectable<Model, Event>

    override func viewDidLoad() {
        let disposable = useMobius(
            controller: controller,
            connectable: viewBinder
        )

        // Disposing the connection will stop the controller.
        // Otherwise it will stop when the current ViewController is deinitialized.
        disposable.dispose()
    }
}
```

@JensAyton 
cc: @togi @pettermahlen

### DIFF
--- a/MobiusExtras/Source/MobiusViewController.swift
+++ b/MobiusExtras/Source/MobiusViewController.swift
@@ -29,9 +29,6 @@ public extension UIViewController {
     /// Note: You should not be calling `start`, `connectView`, `stop`, or `disconnectView` on the `MobiusController`
     /// when using this extension.
     ///
-    /// Note: The `MobiusController` will be placed in a `UIViewController` and added as a child `UIViewController` to
-    ///  this `UIViewController`.
-    ///
     /// - Parameter controller: The `MobiusController` that shuold be attached
     /// - Parameter onModelChange: A closure which is called whenever the loop's model changes.
     func useMobius<Model, Event, Effect>(
@@ -39,21 +36,26 @@ public extension UIViewController {
         modelChanged onModelChange: @escaping (Model) -> Void
     ) -> Connection<Event> {
         let holder = MobiusHolder(controller: controller, onModelChange: onModelChange)
-        addChild(holder)
-        holder.didMove(toParent: self)
+
+        objc_setAssociatedObject(
+            self,
+            .init("Mobius-Controller-Holder-Key"),
+            holder,
+            .OBJC_ASSOCIATION_RETAIN
+        )
+
         return Connection(
             acceptClosure: { [unowned holder] event in
                 holder.handleEvent(event)
             },
             disposeClosure: { [unowned holder] in
-                holder.willMove(toParent: nil)
-                holder.removeFromParent()
+                holder.dispose()
             }
         )
     }
 }
 
-private class MobiusHolder<Model, Event, Effect>: UIViewController {
+private final class MobiusHolder<Model, Event, Effect> {
     private let controller: MobiusController<Model, Event, Effect>
     // swiftlint:disable weak_delegate
     private let connectableDelegate: WeakConnectableDelegate<Model>
@@ -71,24 +73,23 @@ private class MobiusHolder<Model, Event, Effect>: UIViewController {
             }
         )
         self.connectable.delegate = self.connectableDelegate
-        super.init(nibName: nil, bundle: nil)
         controller.connectView(connectable)
         controller.start()
-    }
-
-    required init?(coder: NSCoder) {
-        fatalError("init(coder:) has not been implemented")
     }
 
     func handleEvent(_ event: Event) {
         connectable.send(event)
     }
 
-    deinit {
+    func dispose() {
         if controller.running {
             controller.stop()
             controller.disconnectView()
         }
+    }
+
+    deinit {
+        dispose()
     }
 }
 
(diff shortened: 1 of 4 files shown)


MobiusExtras/Source/MobiusViewController.swift
```diff
@@ -0,0 +1,113 @@
...
+            },
+            disposeClosure: { [unowned holder] in
+                holder.willMove(toParent: nil)
+                holder.removeFromParent()
+            }
+        )
+    }
+}
+
+private class MobiusHolder<Model, Event, Effect>: UIViewController {
```
erikprice: Feels weird, but I suppose I have heard of other cases of using a UIViewController which does not actually control a view. I assume you've already ruled out using a PONSO and `objc_setAssociatedObject()`?

jeppes: That's a good idea, thanks. My initial idea was to use the lifecycle hooks in the child ViewController, but that turned out not to be necessary.

jeppes:
cc: @rastersize

MobiusExtras/Source/MobiusViewController.swift
```diff
@@ -0,0 +1,106 @@
...
+
+#if canImport(UIKit)
+
+import MobiusCore
+import UIKit
+
+public extension UIViewController {
+    /// Attach a `MobiusController` to this `UIViewController`.
+    /// The `connectable` you provide will be connected to the `MobiusController`.
+    /// This function returns a `Disposable` which can be disposed to stop the `MobiusController`. If this disposable
```
rastersize: I think it would be good to document best practices here. My thinking around that would be:

- Call `useMobius` in `viewDidLoad`.
- Dispose the `Disposable` if `didReceiveMemoryWarning` is called.
- The `Disposable` will be automatically disposed when the view controller is disposed.

MobiusExtras/Source/MobiusViewController.swift
```diff
@@ -0,0 +1,106 @@
...
+    /// - Parameter connectable: The `Connectable` which should be connected to `controller`
+    func useMobius<Model, Event, Effect, Conn: Connectable>(
+        controller: MobiusController<Model, Event, Effect>,
+        connectable: Conn
+    ) -> Disposable where Conn.Input == Model, Conn.Output == Event {
+        let holder = MobiusHolder(controller: controller, connectable: AnyConnectable(connectable))
+
+        objc_setAssociatedObject(
+            self,
+            .init("Mobius-Controller-Holder-Key"),
```
JensAyton: Although this is technically correct as far as I can see, putting it in a global/static `let` would make it clearer

MobiusExtras/Source/MobiusViewController.swift
```diff
@@ -0,0 +1,106 @@
...
+public extension UIViewController {
+    /// Attach a `MobiusController` to this `UIViewController`.
+    /// The `connectable` you provide will be connected to the `MobiusController`.
+    /// This function returns a `Disposable` which can be disposed to stop the `MobiusController`. If this disposable
+    /// is never disposed, the `MobiusController` will stop when this `UIViewController` is deinitialized.
+    ///
+    /// Note: You should not be calling `start`, `connectView`, `stop`, or `disconnectView` on the `MobiusController`
+    /// when using this extension.
+    ///
+    /// Note: You must keep a strong reference to the `connectable` for as long as you want the loop to exist.
```
JensAyton: This is a little weird. Even with `WeakConnectable`, as far as I can see in the current design the loop will keep running when the connectable vanishes, it just won’t post view updates?

MobiusExtras/Source/MobiusViewController.swift
JensAyton: `useMobius` can be called in `init` _or_ `viewDidLoad`, but I would tend to prefer `init`. The `MobusController` should not be torn down in `didReceiveMemoryWarning`; maintaining the state you need to recreate it correctly would unnecessarily duplicate state from the `MobiusController` and the model and increase overall memory usage.

MobiusExtras/Source/MobiusViewController.swift
```diff
@@ -0,0 +1,106 @@
...
+    ///
+    /// - Parameter controller: The `MobiusController` that should be attached
+    /// - Parameter connectable: The `Connectable` which should be connected to `controller`
+    func useMobius<Model, Event, Effect, Conn: Connectable>(
```
JensAyton: I think this should be `@discardableResult`. As far as I’m concerned, explicitly disposing this disposable is a weird thing that should rarely happen. (Note that discarding an `AnonymousDisposable` doesn’t trigger it.)

MobiusExtras/Source/MobiusViewController.swift
```diff
@@ -0,0 +1,106 @@
...
+        objc_setAssociatedObject(
+            self,
+            .init("Mobius-Controller-Holder-Key"),
+            holder,
+            .OBJC_ASSOCIATION_RETAIN
+        )
+
+        return AnonymousDisposable {
+            holder.dispose()
+        }
```
JensAyton: Style nit: `holder` should explicitly conform to `Disposable`, and this should be `AnyonymousDisposable(holder)`. It will compile to the same thing, but what we’re expressing here is type erasure, not an action.

jeppes:
Closing for now. Will revisit in the future.

PR #132, opened 2020-03-23
Reorganize error messages

JensAyton:
We’ve seen that stack traces from crashes in `MobiusController` are so deep, our crash analytics merges them into a single report. This PR seeks to flatten those stacks as far as practical.

(We’d be able to get a step further if `ErrorHandler` returned `Never` instead of `Void`, which would also make some call sites nicer. I tried to get us there a while ago, but got stuck because Nimble uses different types for file names depending on the build system in use. I’ll try to revisit that soon.)

(Edit: reading that back I don’t even understand why that was a problem, trying again)

@jeppes

### DIFF
--- a/MobiusCore/Source/AsyncStartStopStateMachine.swift
+++ b/MobiusCore/Source/AsyncStartStopStateMachine.swift
@@ -39,6 +39,10 @@ final class AsyncStartStopStateMachine<StoppedState, RunningState> {
         case transitioningToStopped
     }
 
+    enum Error: Swift.Error {
+        case wrongState
+    }
+
     private let rawState = Synchronized(value: RawState.stopped)
     private let queue: DispatchQueue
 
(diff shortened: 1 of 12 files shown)


MobiusCore/Source/AsyncStartStopStateMachine.swift
```diff
@@ -39,6 +39,10 @@ final class AsyncStartStopStateMachine<StoppedState, RunningState> {
         case transitioningToStopped
     }
 
+    enum Error: Swift.Error {
```
rastersize: How about using a struct instead? Eg

```swift
struct WrongStateError: Error {}

// throw WrongStateError()
```

MobiusCore/Source/MobiusController.swift
```diff
@@ -256,4 +291,25 @@ public final class MobiusController<Model, Event, Effect> {
             }
         }
     }
+
+    /// Simple error that just carries an error message out of a closure for us
+    private enum Error: Swift.Error {
```
rastersize: How about using a struct instead? Would seem more idiomatic to me.

MobiusCore/Source/AsyncStartStopStateMachine.swift
JensAyton: I feel a singleton enum is clearer/less unexpected than a singleton struct for this case, but I’ll change the string wrapper one.

PR #133, opened 2020-03-23
Non returning assertions

JensAyton:
`MobiusHooks.ErrorHandler` now has a return type of `Never`, and so works like a `fatalError` or `preconditionFailure`.

As a result, `BrokenConnection` is no longer needed.

The infrastructure for this is a simple Objective-C library that wraps the Objective-C exception system. (It’s possible to throw an exception from Swift using `NSException.raise()`, but it can’t be caught in pure Swift.)

Nimble already has similar infrastructure, with bells and whistles to automatically catch Swift fatal errors, but that implementation is not thread safe (it triggers TSAN in our CI) and uses a different set of nasty hacks for each runtime environment.

I also use Quick’s `QuickConfiguration` mechanism to install a custom error hook around each test case and remove it afterwards. This is a bit on the magical side, but we had multiple tests that didn’t clean up their error hooks, which meant that the result of tests was order-dependent and one test was passing when it shouldn’t be.

### DIFF
--- a/MobiusCore/Source/BackwardsCompatibility.swift
+++ b/MobiusCore/Source/BackwardsCompatibility.swift
@@ -182,18 +182,6 @@ public extension Connectable {
     typealias OutputType = Output
 }
 
-public extension BrokenConnection {
-    @available(*, deprecated)
-    static func accept(_ value: Value) {
-        _accept(value)
-    }
-
-    @available(*, deprecated)
-    static func dispose() {
-        _dispose()
-    }
-}
-
 @available(*, deprecated)
 public typealias ConnectClosure<InputType, OutputType> = (@escaping Consumer<OutputType>) -> Connection<InputType>
 
--- a/MobiusCore/Test/ConnectablePublisherTests.swift
+++ b/MobiusCore/Test/ConnectablePublisherTests.swift
@@ -1,4 +1,4 @@
-// Copyright (c) 2019 Spotify AB.
+// Copyright (c) 2020 Spotify AB.
 //
 // Licensed to the Apache Software Foundation (ASF) under one
 // or more contributor license agreements.  See the NOTICE file
@@ -29,7 +29,6 @@ class ConnectablePublisherTests: QuickSpec {
             var publisher: ConnectablePublisher<String>!
             var received: [String]!
             var consumer: Consumer<String>!
-            var errorThrown: Bool!
 
             beforeEach {
                 publisher = ConnectablePublisher()
@@ -38,15 +37,6 @@ class ConnectablePublisherTests: QuickSpec {
                 consumer = {
                     received.append($0)
                 }
-
-                errorThrown = false
-                MobiusHooks.setErrorHandler({ _, _, _ in
-                    errorThrown = true
-                })
-            }
-
-            afterEach {
-                MobiusHooks.setDefaultErrorHandler()
             }
 
             describe("connections") {
@@ -138,9 +128,7 @@ class ConnectablePublisherTests: QuickSpec {
 
                     context("when trying to connect") {
                         it("should refuse connections") {
-                            publisher.connect(to: consumer)
-
-                            expect(errorThrown).to(beTrue())
+                            expect(_ = publisher.connect(to: consumer)).to(raiseError())
                         }
 
                         context("when error handling does not cause a crash") {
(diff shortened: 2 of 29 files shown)


MobiusCore/Source/BackwardsCompatibility.swift
```diff
@@ -182,18 +182,6 @@ public extension Connectable {
     typealias OutputType = Output
 }
 
-public extension BrokenConnection {
```
JensAyton: These are moved back into the main implementation, since all of `BrokenConnection` is deprecated now

MobiusCore/Test/ConnectablePublisherTests.swift
```diff
@@ -138,40 +128,13 @@ class ConnectablePublisherTests: QuickSpec {
 
                     context("when trying to connect") {
                         it("should refuse connections") {
-                            publisher.connect(to: consumer)
-
-                            expect(errorThrown).to(beTrue())
-                        }
-
-                        context("when error handling does not cause a crash") {
```
JensAyton: This condition no longer exists

MobiusCore/Test/EffectHandlers/EffectRouterDSLTests.swift
```diff
@@ -253,7 +253,6 @@ class EffectRouterDSLTests: QuickSpec {
                     }
 
                 dslHandler.accept(.effect1)
-                dslHandler.accept(.effect2)
```
JensAyton: This is invalid and caused an error, but it wasn’t visible when running all test cases because some other test’s error handler was left in place

MobiusCore/Test/MobiusControllerTests.swift
```diff
@@ -233,24 +218,20 @@ class MobiusControllerTests: QuickSpec {
                         controller.connectView(view)
                         controller.start()
 
-                        controller.disconnectView()
-                        expect(errorThrown).to(beTrue())
+                        expect(controller.disconnectView()).to(raiseError())
                     }
                 }
 
                 #if arch(x86_64) || arch(arm64)
```
JensAyton: I don’t know why this is here, it predates the public repo. Possibly it was from a previous attempt to use Nimble’s `throwAssertion()` predicate.

MobiusExtras/Test/ConnectableClassTests.swift
```diff
@@ -27,33 +28,42 @@ class ConnectableTests: QuickSpec {
     // swiftlint:disable function_body_length
     override func spec() {
         describe("ConnectableClass") {
-            var errorThrown: Bool = false
-            let handleError = { (_: String) -> Void in
-                errorThrown = true
+            beforeEach {
+                MobiusHooks.setErrorHandler { message, file, line in
+                    MobiusThrowableAssertion(message: message, file: "\(file)", line: line).throw()
+                }
```
JensAyton: Simplified error handler for this one test case in MobiusExtras. Perhaps I’ll look into breaking the testing error handler (and other test utilities?) out of MobiusCoreTests later.

JensAyton:
@jeppes

MobiusCore/Test/MobiusControllerTests.swift
pettermahlen: Yes, I remember there was a need to use this kind of #if statements to only run tests when Nimble supported it.

MobiusCore/Test/TestingErrorHandler.swift
```diff
@@ -0,0 +1,62 @@
+import Foundation
+import MobiusCore
+import MobiusThrowableAssertion
+import Nimble
+import Quick
+
+/// Nimble predicate which passes if the expression invokes `MobiusHooks.effectHandler`.
```
pettermahlen, line 7:
```suggestion
/// Nimble predicate which passes if the expression invokes `MobiusHooks.errorHandler`.
```

MobiusCore/Test/TestingErrorHandler.swift
```diff
@@ -0,0 +1,62 @@
...
+///
+/// The effect handler is overridden before each test in a QuickConfiguration object. If the test has its own override,
```
pettermahlen, line 9:
```suggestion
/// The error handler is overridden before each test in a QuickConfiguration object. If the test has its own override,
```

MobiusCore/Test/TestingErrorHandler.swift
JensAyton, line 7: Oops, this should say `errorHandler`

PR #135, opened 2020-03-25
Alternative Automatic MobiusController lifecycle management in `UIViewController`s

jeppes:
This is an alternative to https://github.com/spotify/Mobius.swift/pull/131

In this version, the MobiusController will be started when the view is visible and stopped when the view disappears.

### DIFF
--- a/MobiusExtras/Test/MobiusViewControllerTests.swift
+++ b/MobiusExtras/Test/MobiusViewControllerTests.swift
@@ -46,7 +46,7 @@ final class MobiusViewControllerTests: QuickSpec {
             it("stops the Mobius Controller when the View Controller's connection is disposed") {
                 let viewController = ViewController(onModelChange: { _ in })
 
-                viewController.connection.dispose()
+                viewController.disposable.dispose()
                 expect(viewController.controller.running).to(beFalse())
                 expect(viewController.children.count).to(equal(0))
             }
@@ -57,9 +57,9 @@ final class MobiusViewControllerTests: QuickSpec {
                     model = newModel
                 })
 
-                viewController.connection.accept("1")
-                viewController.connection.accept("2")
-                viewController.connection.accept("3")
+                viewController.output("1")
+                viewController.output("2")
+                viewController.output("3")
 
                 expect(model).toEventually(equal("123"))
             }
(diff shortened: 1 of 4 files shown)


MobiusExtras/Test/MobiusViewControllerTests.swift
```diff
@@ -0,0 +1,127 @@
...
+
+import Foundation
+import MobiusCore
+import MobiusExtras
+import Nimble
+import Quick
+
+final class MobiusViewControllerTests: QuickSpec {
+    override func spec() {
+        context("View Controller using Mobius") {
```
jeppes: These tests don't actually test that calls to `viewDidAppear` and `viewDidDisappear` are forwarded. I could not think of a way to do this.

rastersize:
@jeppes I’d much prefer the original approach with an ObjC associated object over this approach with a child VC and added subview. Since the latter approach has a higher performance and memory cost. As well as a potential for added confusion when debugging some UI (“where did this view come from” and “erf why is there a view covering all of my UI”). As well as the, while small, added risk that it interferes with some parts of the UI.

jeppes:
@rastersize I agree. I don't know how much value there is in automating the call to stop the loop.
The best solution might be to create a controller similar which just supports starting and stopping and use the approach in #131 to attach it to a controller. Then we could document that it should be started/stopped in viewDidAppear/viewDidDisappear respectively.
It wouldn't be too big of a deal if it weren't stopped, since at least the ViewController would not be leaked.

MobiusExtras/Test/MobiusViewControllerTests.swift
JensAyton: That seems proper, testing that UIKit works as documented isn’t a unit test. :-)

JensAyton:
I feel quite strongly that best practice should be to `start` in `viewWillAppear` and `stop` in `viewDidDisappear`. However, as far as I can see this can’t be automated without messing with the view hierarchy or class swizzling (I can’t see a way to do it with notifications or KVO), and that’s probably too high a cost for a blessed standard pattern.

Right now I’m leaning towards the approach in #131 with documented best practices (and of course we should have a public tutorial which covers it, as well as GP). Note that best practices for a MobiusController that actually uses starty-stoppy behaviour includes having an `initiate` that properly gathers state from any stateful collaborators.

On the other hand, if we do it that way, is it actually any similar than having guidance about how to use `MobiusController` from a VC? One possible benefit I see is that we could add an opt-in protocol that hides the `Connectable` by making `receiveModel` a protocol method and `dispatchEvent` a protocol extension method.

jeppes:
I think we're all in agreement that we don't want this approach, so I will close this PR.

PR #137, opened 2020-03-30
Alternative to #131

jeppes:
This PR proposes an alternative solution to https://github.com/spotify/Mobius.swift/pull/131.

In this approach, we introduce a new type, `MobiusLoopController`, which can be started with a connectable. It should be used in the following way when integrating with a `UIViewController`:

```swift
override func viewDidAppear(...) {
    ...
    controller.start(connectable)
}

override func viewDidDisappear(...) {
    ...
    controller.stop()
}
```

`MobiusLoopController` does not hold a strong reference to its connectable, so if `controller.stop()` is not called it will stop itself when its owner is deinitialized.

If we want to go with this approach, we should definitely discuss the naming, and the migration strategy for `MobiusController`.

@JensAyton @rastersize


jeppes:
Closing for now. Will revisit in the future.

PR #140, opened 2020-04-01
Rename the associated types in the EffectHandler

jeppes:
There has been a lot of confusion about the names of the associated types in the `EffectHandler` protocol, specifically the `Effect` name.
Renaming these to `Input` and `Output` should help with this.

### DIFF
--- a/MobiusCore/Source/EffectHandlers/EffectHandler.swift
+++ b/MobiusCore/Source/EffectHandlers/EffectHandler.swift
@@ -17,18 +17,17 @@
 // specific language governing permissions and limitations
 // under the License.
 
-/// This protocol defines the contract for an Effect Handler which takes `Effect`s as input, and produces `Event`s as
-/// output.
+/// This protocol defines the contract for an Effect Handler which takes `EffectParameters` as input, and produces
+/// `Event`s as output.
 ///
-/// For each incoming `Effect`, zero or more `Event`s can be sent as output using `callback.send(Event)`. When an
-/// `Effect` has been completely handled, i.e. that effect will not result in any more events, call `callback.end()`.
+/// For each incoming effect parameter, zero or more `Event`s can be sent as output using `callback.send(Event)`. When
+/// an the effect parameters have been completely handled, i.e. that effect will not result in any more events, call
+/// `callback.end()`.
 ///
 /// Note: `EffectHandler` should be used in conjunction with an `EffectRouter`.
 public protocol EffectHandler {
-    associatedtype Input
-    associatedtype Output
-    typealias Effect = Input
-    typealias Event = Output
+    associatedtype EffectParameters
+    associatedtype Event
 
     /// Handle an `Effect`.
     ///
@@ -43,27 +42,27 @@ public protocol EffectHandler {
     /// - Parameter input: The effect being handled
     /// - Parameter callback: The `EffectCallback` used to communicate with the associated Mobius loop.
     func handle(
-        _ input: Effect,
+        _ input: EffectParameters,
         _ callback: EffectCallback<Event>
     ) -> Disposable
 }
 
 /// A type-erased wrapper of the `EffectHandler` protocol.
-public struct AnyEffectHandler<Effect, Event>: EffectHandler {
-    private let handleClosure: (Effect, EffectCallback<Event>) -> Disposable
+public struct AnyEffectHandler<EffectParameters, Event>: EffectHandler {
+    private let handleClosure: (EffectParameters, EffectCallback<Event>) -> Disposable
 
     /// Creates an anonymous `EffectHandler` that implements `handle` with the provided closure.
     ///
     /// - Parameter handle: An effect handler `handle` function; see the documentation for `EffectHandler.handle`.
-    public init(handle: @escaping (Effect, EffectCallback<Event>) -> Disposable) {
+    public init(handle: @escaping (EffectParameters, EffectCallback<Event>) -> Disposable) {
         self.handleClosure = handle
     }
 
     /// Creates a type-erased `EffectHandler` that wraps the given instance.
     public init<Handler: EffectHandler>(
         handler: Handler
-    ) where Handler.Effect == Effect, Handler.Event == Event {
-        let handleClosure: (Effect, EffectCallback<Event>) -> Disposable
+    ) where Handler.EffectParameters == EffectParameters, Handler.Event == Event {
+        let handleClosure: (EffectParameters, EffectCallback<Event>) -> Disposable
 
         if let anyHandler = handler as? AnyEffectHandler {
             handleClosure = anyHandler.handleClosure
@@ -74,7 +73,7 @@ public struct AnyEffectHandler<Effect, Event>: EffectHandler {
         self.init(handle: handleClosure)
     }
 
-    public func handle(_ input: Effect, _ callback: EffectCallback<Event>) -> Disposable {
+    public func handle(_ input: EffectParameters, _ callback: EffectCallback<Event>) -> Disposable {
         return self.handleClosure(input, callback)
     }
 }
--- a/MobiusCore/Source/EffectHandlers/EffectHandler.swift
+++ b/MobiusCore/Source/EffectHandlers/EffectHandler.swift
@@ -20,9 +20,9 @@
 /// This protocol defines the contract for an Effect Handler which takes `EffectParameters` as input, and produces
 /// `Event`s as output.
 ///
-/// For each incoming effect parameter, zero or more `Event`s can be sent as output using `callback.send(Event)`. Once
-/// the effect parameters have been completely handled (i.e. when the handler will not produce any more events) call
-/// `callback.end()`.
+/// For each input to the `handle` function, zero or more `Event`s can be sent as output using `callback.send(Event)`.
+/// Once the effect parameters have been completely handled (i.e. when the handler will not produce any more events)
+/// call `callback.end()`.
 ///
 /// Note: `EffectHandler` should be used in conjunction with an `EffectRouter`.
 public protocol EffectHandler {
--- a/MobiusCore/Source/EffectHandlers/EffectHandler.swift
+++ b/MobiusCore/Source/EffectHandlers/EffectHandler.swift
@@ -21,7 +21,7 @@
 /// `Event`s as output.
 ///
 /// For each input to the `handle` function, zero or more `Event`s can be sent as output using `callback.send(Event)`.
-/// Once the effect parameters have been completely handled (i.e. when the handler will not produce any more events)
+/// Once the effect has been completely handled (i.e. when the handler will not produce any more events)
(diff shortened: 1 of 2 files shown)


MobiusCore/Source/EffectHandlers/EffectHandler.swift
```diff
@@ -25,8 +25,10 @@
 ///
 /// Note: `EffectHandler` should be used in conjunction with an `EffectRouter`.
 public protocol EffectHandler {
-    associatedtype Effect
-    associatedtype Event
+    associatedtype Input
+    associatedtype Output
+    typealias Effect = Input
```
jeppes, line 30: We can keep these names around to avoid breaking existing consumers.

laptobbe: 👍 Amazing!

JensAyton:
I haven’t seen this confusion and I’m surprised `Input` and `Output` are seen as more helpful, but if that’s the feedback from users then OK I guess

rastersize:
I’m also confused that `Input` and `Output` are seen as more helpful. Whenever I encounter a Mobius type that uses `Input` and `Output` I have to look at the implementation to figure out what I’m supposed to provide. So for me, at least, this change is a somewhat major day-to-day work regression in usability.

But if the masses wants this then I guess ¯\_(ツ)_/¯

jeppes:
@JensAyton @rastersize It was brought up today again as feedback during the Mobius course, and I've had at least two people mention it to me before that.
The confusion is due to the fact that the `Effect` type is not usually the same as the loop's `Effect` type when using `EffectRouter` - It is instead the type of the associated value of the `Effect` enum for the current case.

JensAyton:
That’s a good point, but I wonder if a pairing like `EffectParameters` (or `EffectState`) and `Event` might be clearer.

I think the `Effect` typealias should be deprecated, there’s still time to get it into 0.3.0.

jeppes:
I like `EffectParameters`.  `Payload` is another option which matches the naming in `_PartialEffectRouter`. Perhaps `EffectPayload`?

JensAyton:
I coined `Payload`, but I think `EffectParameters` is clearer. I don’t think we mention `Payload` anywhere in our codebase yet (even though there’s a `routeEffects(withPayload:)`, so maybe we can change it everywhere now?

jeppes:
I'll change it to `EffectParameters`

JensAyton:
@rastersize Does this make sense to you?

MobiusCore/Source/EffectHandlers/EffectHandler.swift
```diff
@@ -41,27 +42,27 @@ public protocol EffectHandler {
     /// - Parameter input: The effect being handled
     /// - Parameter callback: The `EffectCallback` used to communicate with the associated Mobius loop.
     func handle(
-        _ input: Effect,
+        _ input: EffectParameters,
```
rastersize, line 45: How about naming the argument `effectParameters` or something like that as well?

MobiusCore/Source/EffectHandlers/EffectHandler.swift
```diff
@@ -17,15 +17,16 @@
...
 
-/// This protocol defines the contract for an Effect Handler which takes `Effect`s as input, and produces `Event`s as
-/// output.
+/// This protocol defines the contract for an Effect Handler which takes `EffectParameters` as input, and produces
+/// `Event`s as output.
 ///
-/// For each incoming `Effect`, zero or more `Event`s can be sent as output using `callback.send(Event)`. When an
-/// `Effect` has been completely handled, i.e. that effect will not result in any more events, call `callback.end()`.
+/// For each incoming effect parameter, zero or more `Event`s can be sent as output using `callback.send(Event)`. When
+/// an the effect parameters have been completely handled, i.e. that effect will not result in any more events, call
```
pettermahlen, line 24: This doesn't read well, IMO. In the [Java wiki](https://github.com/spotify/mobius/wiki/Concepts#effect), we talk about effect descriptions and effect execution as separate things. The effect description is the combination of an effect type and values for its parameters. I guess `EffectParameters` to be strict would be the types and names of the parameters (to be analogous with the definition of method parameters), not necessarily their values. In this case, it's parameters with assigned values. Arguments, maybe? Not exactly the same either.

I sort of think that Effect is clearer or at least more precise than EffectParameters. If nothing else, then at least in this paragraph, which should state that "for each incoming effect, zero or more.." rather than "for each incoming effect parameter, ...". The latter sounds like you would potentially send one event for each of the parameters of an effect. 

I do think effect parameters makes sense as a name for the 'names + types' of the data associated with the enum value. But I don't think it makes great sense for the 'names + types + values'. It's not completely unreasonable to use the name `EffectParameters` as a type alias because it's fine to fuzz the distinction about whether there are values or not, but it's definitely not the 'parameters' that the effect handler is handling, but rather the 'effect'.

laptobbe:
For some context, here is an example of how the confusion happens currently. If we go with `EffectParameters` could we also rename the first parameter of handle to `effectParameters`? It's only an internal name so it should not break anything. I think they should match regardless of the name we choose.

```
enum PlaylistCreationEffect: Equatable {
    case presentAlert(title: String, message: String)
}

struct PresentAlertEffectHandler: EffectHandler {
    typealias Effect = PlaylistCreationEffect
    typealias Event = PlaylistCreationEvent

    // Parameter named input of type Effect.
    func handle(_ input: PlaylistCreationEffect, _ callback: EffectCallback<PlaylistCreationEvent>) {
       ...
    }
}

struct PresentAlertEffectHandler: EffectHandler {
    typealias Effect = (String, String)
    typealias Event = PlaylistCreationEvent

    func handle(_ input: (String, String),  _ callback: EffectCallback<PlaylistCreationEvent>) {
       ...
    }
}

```

MobiusCore/Source/EffectHandlers/EffectHandler.swift
jeppes: I have clarified the docs.
I agree that the name is not ideal, but I think it's better than overloading the term `Effect`.

MobiusCore/Source/EffectHandlers/EffectHandler.swift
```diff
@@ -20,29 +20,30 @@
 /// This protocol defines the contract for an Effect Handler which takes `EffectParameters` as input, and produces
 /// `Event`s as output.
 ///
-/// For each incoming effect parameter, zero or more `Event`s can be sent as output using `callback.send(Event)`. When
-/// an the effect parameters have been completely handled, i.e. that effect will not result in any more events, call
+/// For each incoming effect parameter, zero or more `Event`s can be sent as output using `callback.send(Event)`. Once
```
pettermahlen, line 23: This is still weird: "for each incoming effect parameter" - I'm pretty sure most people would think of a parameter set consisting of two Strings as "two effect parameters". 

Better to phrase it as "for each handled Effect", or something like that. Same thing with the sentence below, which makes it sound like each effect parameter (argument! :) ) is handled individually.

MobiusCore/Source/EffectHandlers/EffectHandler.swift
pettermahlen: That's probably true - but the docs actually refer to handling the `Effect` as a whole, not the individual parameters/arguments. I tried to clarify what I mean above.

I'm slightly leaning towards thinking `EffectArguments` is better than `EffectParameters`, but that's not a huge deal, it'll be clear what it means, I think.

JensAyton:
@pettermahlen The parameter/argument distinction isn’t upheld consistently even in formal usages, so I’ve fallen back to using the “formal parameter”/“actual parameter” dichotomy when making the distinction is actually useful. :-)

After understanding the problem yesterday, I feel it’s pretty clear that `Effect` is not the name we want. It is very reasonable to expect that `Effect` means only one thing in the context of a given loop, we shouldn’t use it _both_ for the sum type representing all effects _and_ the arguments to a specific effect. The whole `Event`/`Effect` similarity thing is confusing enough.

pettermahlen:
I think @laptobbe's example clarified to me that the thing in the `PresentAlertEffectHandler` shouldn't be an 'effect'. But the doc comment shouldn't talk about handling the parameters - it's handling the effect through a method that accepts the parameters. We may be saying the same thing here. :)

Trying to figure out how I think about this: 'Effect' is currently overloaded to mean 'effect description' and 'effect execution', but we usually mean 'effect description'. 'Effect description' means 'effect type' + parameter values. In this case, the typealias in the `PresentAlertEffectHandler` refers to the effect parameters (types + names), and the parameter to the 'handle' function is really the same thing? So maybe I'm coming around to thinking that parameters is an OK name for it. The doc comment should still talk about handling the effect, though, IMO. It's true that only the parameters are passed to the 'handler' function, but the parameters aren't what's being handled.

MobiusCore/Source/EffectHandlers/EffectHandler.swift
```diff
@@ -17,51 +17,53 @@
...
 
-/// This protocol defines the contract for an Effect Handler which takes `Effect`s as input, and produces `Event`s as
-/// output.
+/// This protocol defines the contract for an Effect Handler which takes `EffectParameters` as input, and produces
+/// `Event`s as output.
 ///
-/// For each incoming `Effect`, zero or more `Event`s can be sent as output using `callback.send(Event)`. When an
-/// `Effect` has been completely handled, i.e. that effect will not result in any more events, call `callback.end()`.
+/// For each input to the `handle` function, zero or more `Event`s can be sent as output using `callback.send(Event)`.
+/// Once the effect parameters have been completely handled (i.e. when the handler will not produce any more events)
```
pettermahlen, line 24:
```suggestion
/// Once the effect has been completely handled (i.e. when the handler will not produce any more events)
```

MobiusCore/Source/EffectHandlers/EffectHandler.swift
```diff
@@ -72,7 +74,7 @@ public struct AnyEffectHandler<Effect, Event>: EffectHandler {
         self.init(handle: handleClosure)
     }
 
-    public func handle(_ input: Effect, _ callback: EffectCallback<Event>) -> Disposable {
+    public func handle(_ input: EffectParameters, _ callback: EffectCallback<Event>) -> Disposable {
```
pettermahlen, line 77:
```suggestion
    public func handle(_ parameters: EffectParameters, _ callback: EffectCallback<Event>) -> Disposable {
```

MobiusCore/Source/EffectHandlers/EffectHandler.swift
```diff
@@ -72,7 +74,7 @@ public struct AnyEffectHandler<Effect, Event>: EffectHandler {
...
         return self.handleClosure(input, callback)
```
pettermahlen, line 78:
```suggestion
        return self.handleClosure(parameters, callback)
```

PR #141, opened 2020-04-02
[WIP] Improved equatability errors in Swift 5.2 when using the EffectRouter

jeppes:
Using the `EffectRouter` with non-equatable enums in Swift 5.2 results in very unhelpful error messages. This PR tries to resolve that issue.

Pre Swift 5.2:
pre-swift-5_2
Swift 5.2:
swift-5_2

After this change we are back to the correct error message:
swift-5_2-post-change


jeppes:
It turns out this is very difficult to do well 🙃

PR #143, opened 2020-04-03
Bump 0.3.0

JensAyton:
0.3.0 makes many changes from 0.2.0. Where possible, old names and types are available with deprecation attributes; these will soon be removed.

- Updated threading model:
  - `MobiusLoop` is now single-threaded
  - `MobiusController` runs a loop on a single background queue.
  - Fixed several issues around hard-to-avoid assertions in loop teardown.
- New `EffectRouter` and `EffectHandler` replace `EffectRouterBuilder` (which is deprecated along with several helpers).
- Effects in `First` and `Next` are now an array rather than a set. This doesn’t imply an ordering guarantee, but does mean that effects don’t have to be `Hashable`.
- Various things renamed or changed from methods to properties to better conform to Swift API Guidelines and for internal consistency:
  - `Initiator` is now `Initiate`, and is only used with `MobiusController`.
  - `MobiusLoop.getMostRecentModel()` becomes `latestModel`.
  - `MobiusController.getModel()` becomes `model`.
  - `Connectable.InputType` and `OutputType` become `Input` and `Output`; `Connection.ValueType` becomes `Value`.
- `Update` is now a struct. This isn’t leveraged by Mobius itself at the moment, but makes it easier to write transformations on updates in a fluent style.
- For consistency, `MobiusController` is created through a `makeController()` method on `Mobius.Builder` instead of being initialized with a builder argument.
- All methods on `MobiusLogger` now have default do-nothing implementations.
- `ConsoleLogger` has been replaced with `SimpleLogger`, which can take a consumer function to use instead of `print`.
- `NoEffect` and `BrokenConnection` are deprecated.
- `MobiusHooks.ErrorHandler` now returns `Never` rather than `Void`.
- Mobius no longer adds a public extension to `NSRecursiveLock`.
- There are more documentation comments than there used to be.
- Tooling updated to Swift 5.0 and Xcode 11.0.
- Swift Package Manager is explicitly supported for all Apple platforms; Carthage and CocoaPods are supported for iOS only.

### DIFF
--- a/MobiusCore.podspec.json
+++ b/MobiusCore.podspec.json
@@ -1,6 +1,6 @@
 {
   "name": "MobiusCore",
-  "version": "0.2.0",
+  "version": "0.3.0",
   "summary": "A functional reactive framework for managing state evolution and side-effects",
   "authors": "Spotify AB",
   "homepage": "https://github.com/spotify/Mobius.swift",
@@ -11,7 +11,7 @@
   },
   "source": {
     "git": "https://github.com/spotify/Mobius.swift.git",
-    "tag": "0.2.0"
+    "tag": "0.3.0"
   },
   "platforms": {
     "ios": "10.0"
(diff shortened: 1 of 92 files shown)


JensAyton:
On second thought, I think we want #144

PR #144, opened 2020-04-03
Deprecate old EffectRouter options

JensAyton:
It occurred to me at the last minute that we want to do this for 0.3.0. Internally, we’ll need to bump to the version before this until we finish the migration, but we want to drop deprecated stuff on master.

@jeppes


JensAyton:
So actually, this was left unmerged so I could discuss it with @jeppes next week.

JensAyton:
Discussion done in post

PR #151, opened 2020-04-20
Simplify MobiusLoop implementation

JensAyton:
A complete rewrite of `MobiusLoop`, without `EventProcessor`.

After the threading rewrite, `EventProcessor` was just a wordy way to deal with initialization order and queue up effects received during setup. The latter can be handled by the `WorkBag`, and while the initialization dependencies are non-trivial, merging three layers of code into one makes things simpler.

Behaviour change: effects passed to the initializer or posted in event source/effect handler subscription are now handled in randomized order. Previously they were queued up deterministically in `EventProcessor`, which was an oversight.

@jeppes @pettermahlen


MobiusCore/Source/MobiusLoop.swift
```diff
@@ -19,39 +19,77 @@
...
+        // be a closure so that it can be transformed by eventConsumerTransformer, and to handle ownership correctly:
+        // consumeEvent holds on to the update function and workbag, and also holds self while its work bag submission
+        // is queued.
+        //
+        // Originally the processNext(...) invocation was wrapped in a method, but that just spread things out more.
+        let consumeEvent = eventConsumerTransformer { [unowned self] event in
+            // Note: captures self strongly until the block is serviced by the workBag
+            let processNext = self.processNext
+            workBag.submit {
+                processNext(loggingUpdate.update(model: self.model, event: event))
```
pettermahlen, line 67: This is hard for me to parse - but I hope it captures the value of `self.model` at actual execution time, not at 'event consume' time. That is, it must be the case that event E2 is processed with the model resulting from processing event E1, whether E2 is submitted before or after E1 has finished processing. Is that correct?

MobiusCore/Source/MobiusLoop.swift
```diff
@@ -19,39 +19,77 @@
...
-    private var disposed = false
-    private var access: ConcurrentAccessDetector
+/// A `MobiusLoop` is the core encapsulation of business logic in Mobius.
+///
+/// It stores a current model, applies incoming events, processes the resulting effects, and broadcasts model changes
+/// to observers.
+///
+/// Use `Mobius.loop(update:effectHandler:)` to create an instance.
+public final class MobiusLoop<Model, Event, Effect>: Disposable {
+    private var access = ConcurrentAccessDetector()
```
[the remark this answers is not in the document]
JensAyton: Technically not, but morally `ConcurrentAccessDetector.guard` should be marked `mutating`

MobiusCore/Source/MobiusLoop.swift
```diff
@@ -19,39 +19,77 @@
...
+        self.modelPublisher = ConnectablePublisher<Model>()
+
+        // consumeEvent is the closure that processes an event and handles the model and effect updates. It needs to
+        // be a closure so that it can be transformed by eventConsumerTransformer, and to handle ownership correctly:
+        // consumeEvent holds on to the update function and workbag, and also holds self while its work bag submission
+        // is queued.
+        //
+        // Originally the processNext(...) invocation was wrapped in a method, but that just spread things out more.
+        let consumeEvent = eventConsumerTransformer { [unowned self] event in
+            // Note: captures self strongly until the block is serviced by the workBag
```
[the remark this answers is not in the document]
JensAyton: Yes.

MobiusCore/Source/MobiusLoop.swift
JensAyton: Yes, this is correct. If we had `let model = self.model` outside the `workBag.submit` and used it here, we would hit the issue you’re describing (and I can verify that existing unit tests fail in that scenario). As written, `self.model` is read immediately before calling `processNext` (this is an unowned read of `self`, but the lifetime is guaranteed by the curried `processNext`).

JensAyton: I’ll add a comment to clarify this

MobiusCore/Source/MobiusLoop.swift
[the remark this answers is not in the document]
JensAyton: However, it’s `let` in other places, so I’ll make it `let` and then change it everywhere

JensAyton:
The redness of the build is from a non-mandatory codecov -0.13% delta

MobiusCore/Source/MobiusLoop.swift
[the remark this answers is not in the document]
jeppes: Right, I was also suprised that `ConcurrentAccessDetector.guard` wasn't `mutating`

MobiusCore/Source/MobiusLoop.swift
JensAyton: Comment added BTW

pettermahlen: Looks good!

PR #153, opened 2020-04-20
Fix and simplify logging adaptors

JensAyton:
Through a series of unfortunate events, `Update.logging` was refactored to a point where it didn’t actually work as far as the stack annotations are concerned.

This implementation works, and is simpler and more uniform.

@jeppes


JensAyton:
Red because it reduces coverage by removing unnecessarily-long but well covered code

PR #154, opened 2020-04-20
Misc. cleanup

JensAyton:
Odds and ends:

* EventSource+Extensions.swift is now EventSourceMap.swift for consistency with ConnectableContramap/ConnectableMap
* `AnyConnectable` is now a struct, since it has no reason not to be
* `Mobius` enum is declared the normal way
* Default error handler no longer prints call stack, default behaviour should be as vanilla as possible and clients can customize already
* `BeginnerLoop` uses `EffectRouter` to avoid guiding people in the wrong direction

@jeppes


JensAyton:
Fixed a silly bug. I’m surprised that was possible!

PR #155, opened 2020-04-21
Proposal: Add ability to specify execution queue in `EffectRouter`

jeppes:
This adds the ability to specify which queue an effect should be handled on in the `EffectRouter` with the following syntax:
```swift
EffectRouter<Effect, Event>()
    .routeCase(...)
    .on(.main) // <--- execute this effect on the main thread
    .to { /* some UI action */ }
```
Previously, this would have required writing:
```swift
EffectRouter<Effect, Event>()
    .routeCase(...)
    .to { 
        DispatchQueue.main.async { 
            // Some UI action
        }
    }
```

This can safely be used in both raw loops and in `MobiusController` loops.

@JensAyton @pettermahlen


MobiusCore/Source/EffectHandlers/EffectRouter.swift
```diff
@@ -98,24 +101,40 @@ public struct _PartialEffectRouter<Effect, EffectParameters, Event> {
...
         let connectable = ThreadSafeConnectable(connectable: connectable)
-        let route = Route(extractParameters: path, connectable: connectable)
+        let route = Route(extractParameters: path, connectable: connectable, queue: queue)
         return EffectRouter(routes: routes + [route])
     }
+
+    /// Handle an the current `Effect` asynchronously on the provided `DispatchQueue`
+    ///
+    /// - Parameter queue: The `DispatchQueue` that the current `Effect` should be handled on.
+    public func on(_ queue: DispatchQueue) -> Self {
```
jeppes, line 111: This is currently returning `Self`, which means that `.on(..)` can be called multiple times on the same route.

This could be solved by adding another intermediate type. I'm not sure this is worth it.

We could also crash if a `queue` has already been set. Again, I think the risk of this is quite low, but we can discuss this further here in the comments.

PR #160, opened 2020-04-30
Report errors in user files instead of in MobiusController

jeppes:
@JensAyton


jeppes:
`CI / spm` seems to be consistently failing due to a module cache issue. Any ideas on how we can fix this @JensAyton?

JensAyton:
I’ve been seeing this issue but just retriggering because I’m lazy and I don’t know anything about the CI setup.

@dflems Could you give us some guidance on how to debug this?

jeppes:
@dflems I retriggered this one, but here is a link to a failing run: https://github.com/spotify/Mobius.swift/pull/160/checks?check_run_id=632567573

dflems:
@JensAyton @jeppes: I think #162 should take care of it.

JensAyton:
I don’t actually think this is the Right Thing. Reporting precondition violations inside the code that detects them is normal behaviour, and the feedback we got was that the messages were actionable but it was confusing because our crash reporter aggregated a bunch of unrelated issues. Changing the file and line in the error message won’t affect the stack trace and won’t fix the aggregation issue.

jeppes:
You are right that it will not fix the aggregation issue.

However, I don't think `cannot stop a controller while not running, [...]/MobiusCore/Source/MobiusController.swift:253` is preferable to: `cannot stop a controller while not running, [...]/SomeViewController.swift:253`.

If we do go with the former, I would prefer if we removed the `[...]/MobiusCore/Source/MobiusController.swift:253` part since it's not meaningful to the person reading the error message and might even be misleading.

If we go with the latter, perhaps we should change the message to something like: `Cannot stop a controller while not running. Controller was stopped in: [...]/SomeViewController.swift:253`.

JensAyton:
By this logic, _all_ APIs should take file and line parameters. I don’t see any clear reason why Mobius should be special in this regard.

jeppes:
No, that's not at all what this logic suggests. Not all APIs crash. Even if they did, not all APIs crash because the user violated a precondition. Even if that was the case, not all of those cases could be directly tied to a single line written by the user.

In this case, however, we can very clearly tie the crash directly to a single line in the API consumer's code. But instead of making that information immediately accessible, we instead present an unhelpful file path and line number.

jeppes:
To avoid breaking the convention, we could report the following:
`controller while not running. Controller was stopped in: [...]/SomeViewController.swift:253, [...]/MobiusCore/Source/MobiusController.swift:253`.
Given that this does not affect the consumer's code, I don't think we should omit the line and file where `.stop` was called. This feels like we are actively omitting useful information.

That being said, I would prefer if we omitted the `[...]/MobiusCore/Source/MobiusController.swift:253` part, since this makes it feel like we are actively conveying misleading information.

pettermahlen:
If the information about where in the code the error was made doesn't make it to the stack trace automatically, I would agree with @jeppes. If the stack trace includes the place where Mobius is called, then I don't think we need to report the file+line inside Mobius at all?

JensAyton:
The stack trace does include the call site. The file and line logging ultimately happens in Spotify’s assertion functions, not Mobius.

jeppes:
I will close this. It would be great if we could find a less intrusive way of doing this.

PR #161, opened 2020-04-30
Connectable class is no longer abstract

jeppes:
Trying to make `ConnectableClass` into an abstract class has not worked out well in practice. Most subclasses simply just leave `onDispose` blank. 

@JensAyton


JensAyton:
This one has failing tests, not flakiness

JensAyton:
Otherwise good

jeppes:
Right. 🤦

jeppes:
Xcode's error color really doesn't play well with my color blindness 🙃

jeppes:
Fixed, @JensAyton

PR #164, opened 2020-05-25
Visual model diffing in MobiusTest

zvonicek:
_This PR is a draft to see what are the thoughts about this. If the sentiment is positive I'll finalize the code and write unit tests for it._

The PR implements visual model diffing to MobiusTest. It makes it easy to see what the difference was if the expected model is not matching the actual one. The feature got inspired by [visual diff](https://twitter.com/stephencelis/status/1263600211273924610) in Point-Free Composable Architecture and the experience with my team's current project, where we operate with quite large models and seeing why exactly the spec test failed was always hard there.

One difference compared to the Composable Architecture is that they use their own prettier dump printer, but I found the Swift's `dump` good enough.

Before:
Screenshot 2020-05-25 at 21 33 02

After:
Screenshot 2020-05-25 at 21 31 47

The diff algorithm is adopted and modified from https://github.com/paulgb/simplediff that is licensed under zlib/libpng.


MobiusTest/Source/FirstMatchers.swift
```diff
@@ -53,7 +53,8 @@ public func hasModel<Model: Equatable, Effect>(
     return { (first: First<Model, Effect>) in
         if first.model != expected {
             return .failure(
-                message: "Expected model to be <\(expected)>, got <\(first.model)>",
+                message: "Expected model to be different (−) than what got (+): \n" +
```
pettermahlen, line 56:
```suggestion
                message: "Different model than expected (−), got (+): \n" +
```
The current message isn't grammatically correct. Something like the suggested version is also a bit shorter.

MobiusTest/Source/NextMatchers.swift
```diff
@@ -94,8 +94,8 @@ public func hasModel<Model: Equatable, Effect>(
         let actual = next.model
         if actual != expected {
             return .failure(
-                message: "Expected final Next to have model: <\(String(describing: expected))>. " +
-                    "Got: <\(String(describing: actual))>",
+                message: "Expected final Next model to be different (−) than what got (+): \n" +
```
pettermahlen, line 97:
```suggestion
                message: "Different final model than expected (−), got (+): \n" +
```

zvonicek:
Thanks for the comments @pettermahlen! I like the suggested error messages better.

I'll wait a bit if @JensAyton, @jeppes or others have any thoughts and will try to finalize the PR then.

MobiusTest/Source/DebugDiff.swift
```diff
@@ -0,0 +1,112 @@
...
+          return "+"
+        case .delete:
+          return "−"
+        case .same:
+          return "\u{2007}"
+        }
+    }
+}
+
+// Adopted from https://github.com/paulgb/simplediff
```
JensAyton, line 71: This should be in its own file, with a [zlib license header](https://github.com/paulgb/simplediff/blob/master/LICENSE).

jeppes:
This looks great! I think you should move forward with finalizing the PR

zvonicek:
Thanks, I'll dedicate it some time this week.

kmcbride:
Nice work! It looks like you just need to add `DebugDiff.swift` and `SimpleDiff.swift` to the `MobiusTest-static-ios` target in order to get the build passing 🙂

zvonicek:
Thanks Kylan @kmcbride, I was wondering what the problem is. 

@pettermahlen I covered the diff function with tests and included the case for identical inputs, but I'm not sure if it's also worth testing the entire message in `NextMatchersTests.swift` and `FirstMatchersTests.swift` as that would expose the dump output format there.

MobiusTest/Source/DebugDiff.swift
```diff
@@ -0,0 +1,42 @@
...
+    let rhsLines = dumpUnwrapped(rhs).split(separator: "\n")
+
+    let diffList = diff(lhs: ArraySlice(lhsLines), rhs: ArraySlice(rhsLines))
+
+    return diffList.flatMap { diff in
+        diff.string.map { "\(diff.prefix)\($0)" }
+    }.joined(separator: "\n")
+}
+
+func dumpUnwrapped<T>(_ value: T) -> String {
```
jeppes, line 31: `dumpUnwrap` should be private and not tested in my opinion.

zvonicek: I agree, I can also test those cases (diff input is Optional or nil) directly using `dumpDiff`. I'll change it accordingly.

PR #169, opened 2020-10-21
Remove strong connectable<->loop reference

kmcbride:
This addresses an issue with the loop being retained by every instance that holds a reference to the `consumer` that is passed into the view connectable. It replaces the function reference with an indirect, unowned loop reference.

I've included a regression test case to demonstrate the problem/fix.

### DIFF
--- a/MobiusCore/Source/MobiusController.swift
+++ b/MobiusCore/Source/MobiusController.swift
@@ -216,9 +216,11 @@ public final class MobiusController<Model, Event, Effect> {
                 var disposables: [Disposable] = [loop]
 
                 if let viewConnectable = stoppedState.viewConnectable {
-                    // Note: loop.unguardedDispatchEvent will call our flipEventsToLoopQueue, which implements the assertion
-                    // that “unguarded” refers to, and also (of course) flips to the loop queue.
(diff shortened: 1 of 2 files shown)


pettermahlen:
Isn't this a problem in how the `viewConnection` is disposed? That is, it should release the reference to the consumer/loop when disposed? 

If that is correct, I guess that logic lives in userland, outside of the framework, and therefore that a solution like this makes it easier for users not to make a mistake?

kmcbride:
@pettermahlen that is an accurate assessment. This isn’t fixing a bug with Mobius itself, but more-so adding an extra layer of protection for when users mistakenly keep a reference to the consumer without realizing the implications.

JensAyton:
In general, we’ve aimed to work as expected when things are destroyed without being disposed, which is in line with general Swift expectations. My main concern is whether this could break existing code, but it seems that would require that code to rely on a loop running when there are no explicit references to the loop itself, which seems unlikely.

I suggest we accept this and bump the version to 0.4.0, but I’m not clear on wether @pettermahlen was implying an objection or not.

pettermahlen:
No objections from me! :)

JensAyton (approved):

PR #173, opened 2020-12-04
WIP: Diff effects in NextPredicate

JensAyton:
Extend #164 to missing effects.

Because the expected effects are unordered and unexpected effects are allowed, it uses fuzzy matching to guess which actual effect is closest to a given missing effect.

TODO: unit tests.


kmcbride:
Closing as a result of these changes being rolled into https://github.com/spotify/Mobius.swift/pull/177

PR #176, opened 2021-04-30
Support multiple controller view connections

kmcbride:
Following up on an idea that had been proposed.

### DIFF
--- a/MobiusCore/Source/MobiusController.swift
+++ b/MobiusCore/Source/MobiusController.swift
@@ -10,17 +10,15 @@ import Foundation
 public final class MobiusController<Model, Event, Effect> {
     typealias Loop = MobiusLoop<Model, Event, Effect>
     typealias ViewConnectable = AsyncDispatchQueueConnectable<Model, Event>
-    typealias ViewConnection = Connection<Model>
 
     private struct StoppedState {
         var modelToStartFrom: Model
-        var viewConnectable: ViewConnectable?
-
+        var viewConnectables: [UUID: ViewConnectable]
     }
 
     private struct RunningState {
         var loop: Loop
-        var viewConnectable: ViewConnectable?
+        var viewConnectables: [UUID: ViewConnectable]
         var disposables: CompositeDisposable
     }
 
@@ -78,7 +76,7 @@ public final class MobiusController<Model, Event, Effect> {
         self.viewQueue = viewQueue
 
         let state = State(
-            state: StoppedState(modelToStartFrom: initialModel, viewConnectable: nil),
+            state: StoppedState(modelToStartFrom: initialModel, viewConnectables: [:]),
             queue: loopQueue
         )
         self.state = state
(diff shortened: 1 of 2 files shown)


MobiusCore/Source/MobiusController.swift
```diff
@@ -149,6 +147,38 @@ public final class MobiusController<Model, Event, Effect> {
...
+    /// Model updates will be sent to the connection on the provided `DispatchQueue`.
+    ///
+    /// - Parameter connectable: The `Connectable` to connect upon loop start.
+    /// - Parameter acceptQueue: The `DispatchQueue` on which the connection will receive input.
+    /// - Returns: A `Disposable` that can be used to remove the `Connectable`.
+    /// - Attention: Fails via `MobiusHooks.errorHandler` if the loop is running.
+    @discardableResult
+    public func connect<LoopConnectable: Connectable>(
+        _ connectable: LoopConnectable,
+        acceptQueue: DispatchQueue
```
rastersize, line 162: Would it make sense to use `.main` as a default for this? My guesstimate is the vast majority of usage is connecting to UI where `.main` is required. For other uses cases I’d say `.main` is likely what should be used until there’s a measurable performance issue, or known threading issue.

MobiusCore/Source/MobiusController.swift
```diff
@@ -303,6 +334,29 @@ public final class MobiusController<Model, Event, Effect> {
...
+                errorMessage(error, default: "\(Self.debugTag): cannot disconnect while running; call stop first"),
+                #file,
+                #line
+            )
+        }
+    }
+
+    private static var viewConnectableID: UUID {
+        // Reserve 'nil' to prevent collision with generated connection ids
+        return UUID(uuid: (0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0))
```
rastersize, line 357: Is this really necessary? Given that `UUID` is documented as (emphasis mine):

> A _universally unique_ value that can be used to identify types, interfaces, and other items.

I guess it doesn’t hurt, but?

rastersize:
I like this idea 👍

MobiusCore/Source/MobiusController.swift
kmcbride: 👍 That seems like a reasonable assumption.

MobiusCore/Source/MobiusController.swift
kmcbride: Yeah, it appears unnecessary but the reason behind it is that static stored properties aren't supported in types that use generics. This needs to return a consistent value, so I figured I might as well use [one](https://tools.ietf.org/html/rfc4122#section-4.1.7) that _shouldn't_ be generated otherwise.

rastersize: Thanks for the explanation. Makes sense to me, I'm good with keeping it personally.

MobiusCore/Source/MobiusController.swift
```diff
@@ -130,25 +128,24 @@ public final class MobiusController<Model, Event, Effect> {
 
     /// Connect a view to this controller.
     ///
-    /// May not be called while the loop is running.
-    ///
```
zvonicek, line 134: Isn't this still true even after the change?

zvonicek (approved):

PR #177, opened 2021-05-01
Diff effects in NextPredicate + tests

zvonicek:
This PR extends existing @JensAyton's draft PR #173 with tests as an aim to productionize it. The intention was to make as minimal changes on top of #173 as possible, but I ended up making two:
- Modified the error message with diff signs explanation, which helps to interpret the error and brings the message closer to the existing one in `hasModel` (62651c7ede15462d00fb0666b3691fcee047f447) 
- Prefixed dump of a missing effect with no closest match with `-` sign to indicate that the effect is missing (f79fd3ba2850887dc1fbf72b1994570c913e7583)

I also discovered one limitation of the implementation that I did not fix (yet?): 

- If one unmatched effect has multiple possible candidates in the expected effects, the diff against the unmatched effect is used on all those candidates repeatedly (visible also on the screenshot below). This does not correspond to how diffing typically works as the chunk shouldn't be used repeatedly.

----------

obrazek

### DIFF
--- a/MobiusTest/Source/DebugDiff.swift
+++ b/MobiusTest/Source/DebugDiff.swift
@@ -45,9 +45,10 @@ func closestDiff<T, S: Sequence>(
     for value: T,
     in sequence: S,
     predicate: ([Difference]) -> Bool = { _ in true }
-) -> [Difference]? where S.Element == T {
-    var closest: [Difference]?
+) -> ([Difference]?, T?) where S.Element == T {
+    var closestDiff: [Difference]?
     var closestDistance = Int.max
+    var closestCandidate: T?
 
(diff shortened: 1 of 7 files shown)


kmcbride:
I'm in favor of the suggestion to not reuse already matched diffs. Looks good overall.

zvonicek:
Thanks @kmcbride, I fixed the diff reuse issue. The example from above now looks like this:

Screenshot 2021-05-09 at 17 20 32

zvonicek:
I finally got back to this PR and addressed the comment from @JensAyton. The result is a generalized `dumpDiffFuzzy` function that keep as much diffing functionality as possible away from the matchers. I also added support for the new effect matchers that @kmcbride implemented last week.

kmcbride (approved):

PR #178, opened 2021-08-22
Add more specific effect matchers

kmcbride:
Introduces a few additional effect matchers to `MobiusTest` and `MobiusNimble`:
- `hasOnlyEffects` / `haveOnlyEffects`: Enforces that _only_ the expected effects are present, regardless of order
- `hasExactEffects` / `haveExactEffects`: Enforces that _only_ the expected effects are present, in the specified order

### DIFF
--- a/MobiusTest/Source/NextMatchers.swift
+++ b/MobiusTest/Source/NextMatchers.swift
@@ -132,9 +132,48 @@ public func hasEffects<Model, Effect: Equatable>(
     line: UInt = #line
 ) -> NextPredicate<Model, Effect> {
     return { (next: Next<Model, Effect>) in
-        let actual = next.effects
-        if !expected.allSatisfy(actual.contains) {
-            return .failure(message: "Expected <\(actual)> to contain <\(expected)>", file: file, line: line)
+        if !expected.allSatisfy(next.effects.contains) {
+            return .failure(message: "Expected <\(next.effects)> to contain <\(expected)>", file: file, line: line)
+        }
+        return .success
+    }
+}
+
+/// Constructs a matcher that matches if only the supplied effects are present in the supplied `Next`, in any order.
+///
+/// - Parameter expected: the effects to match (possibly empty)
+/// - Returns: a `Predicate` that matches `Next` instances that include all the supplied effects
+public func hasOnlyEffects<Model, Effect: Equatable>(
+    _ expected: [Effect],
+    file: StaticString = #file,
+    line: UInt = #line
+) -> NextPredicate<Model, Effect> {
+    return { (next: Next<Model, Effect>) in
+        var unmatchedActual = next.effects
+        var unmatchedExpected = expected
+        zip(next.effects, expected).forEach {
+            _ = unmatchedActual.firstIndex(of: $1).map { unmatchedActual.remove(at: $0) }
+            _ = unmatchedExpected.firstIndex(of: $0).map { unmatchedExpected.remove(at: $0) }
+        }
+        if !unmatchedActual.isEmpty || !unmatchedExpected.isEmpty {
+            return .failure(message: "Expected <\(next.effects)> to contain only <\(expected)>", file: file, line: line)
+        }
+        return .success
+    }
+}
+
+/// Constructs a matcher that matches if the supplied effects are equal to the supplied `Next`.
+///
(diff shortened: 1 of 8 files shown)


rastersize (approved):
Is it just me or is it a bit weird that the Nimble matchers are named `have` while the MobiusTest matchers are named `has`? Not a big deal but seems like it would make sense to use the same prefix.

For a future change I wonder if we should rename (as-in add a new method named that, migrate existing code to it and then remove the old method) `(has|have)Effects` to `containsEffect`? In order to make it more clear which matcher does what.

kmcbride:
> Is it just me or is it a bit weird that the Nimble matchers are named `have` while the MobiusTest matchers are named `has`? Not a big deal but seems like it would make sense to use the same prefix.

I initially thought that as well, but it helps tests read more naturally. I assume that's the original reasoning for it.
- Standard: `assertThatNext(hasOnlyEffects([.foo, .bar]))`
- Nimble: `expect(next).to(haveOnlyEffects([.foo, .bar]))`

> For a future change I wonder if we should rename (as-in add a new method named that, migrate existing code to it and then remove the old method) `(has|have)Effects` to `containsEffect`? In order to make it more clear which matcher does what.

Yeah, I struggled with how to best name them. The problem is that `contains` doesn't seem to have a consistent meaning across frameworks — in Hamcrest it means "exactly these items in this order" but in Nimble it's "at least these items in any order".

rastersize:
> > Is it just me or is it a bit weird that the Nimble matchers are named `have` while the MobiusTest matchers are named `has`? Not a big deal but seems like it would make sense to use the same prefix.
> 
> I initially thought that as well, but it helps tests read more naturally. I assume that's the original reasoning for it.
> 
> * Standard: `assertThatNext(hasOnlyEffects([.foo, .bar]))`
> * Nimble: `expect(next).to(haveOnlyEffects([.foo, .bar]))`

Ah yeah that makes sense 👍 

> > For a future change I wonder if we should rename (as-in add a new method named that, migrate existing code to it and then remove the old method) `(has|have)Effects` to `containsEffect`? In order to make it more clear which matcher does what.
> 
> Yeah, I struggled with how to best name them. The problem is that `contains` doesn't seem to have a consistent meaning across frameworks — in Hamcrest it means "exactly these items in this order" but in Nimble it's "at least these items in any order".

I think Hamcrest is the odd one out here. As both Nimble and the Swift standard library uses `contains` as “has at least these items”. However I feel like such a change might warrant more input from the wider (internal at least) iOS community.

pettermahlen:
I think both 'has' and 'contains' mean the same thing - that is, no difference regarding whether it means 'has at least' or 'has exactly' or 'contains at least' and 'contains exactly'. So for precision, a longer name is warranted, in my opinion.

MobiusTest/Source/NextMatchers.swift
```diff
@@ -132,9 +132,48 @@ public func hasEffects<Model, Effect: Equatable>(
...
+    file: StaticString = #file,
+    line: UInt = #line
+) -> NextPredicate<Model, Effect> {
+    return { (next: Next<Model, Effect>) in
+        var unmatchedActual = next.effects
+        var unmatchedExpected = expected
+        zip(next.effects, expected).forEach {
+            _ = unmatchedActual.firstIndex(of: $1).map { unmatchedActual.remove(at: $0) }
+            _ = unmatchedExpected.firstIndex(of: $0).map { unmatchedExpected.remove(at: $0) }
+        }
```
zvonicek, line 157: @kmcbride I'm trying to finalize #177 (finally!) and this code is causing me troubles. When for example `next.effects == [.a, .b]` and `expected == [.b]`, then after running this code `unmatchedExpected == [.b]`, which I think is incorrect, as `.b` could actually be matched.

Could this be implemented this way (which would in my opinion fix the problem):
```
let unmatchedActual = next.effects.filter { !expected.contains($0) }
let unmatchedExpected = expected.filter { !next.effects.contains($0) }
```

or is there any edge case I'm not seeing here?

kmcbride: The intention here was to do the minimum work necessary to determine if the expectation could be satisfied, so the arrays don't always contain the "true" diff between the two inputs even though the returned value is correct. Feel free to adjust as needed for computing the actual diff 😬

PR #180, opened 2021-09-04
Update minimum dependency versions

kmcbride:
Fixes an issue with bootstrap failing when building with Xcode 13.


rastersize (approved):
Thank you!

PR #184, opened 2022-01-02
Fix CasePath object-constrained existential extraction

kmcbride:
Ports the [fix](https://github.com/pointfreeco/swift-case-paths/pull/65) for the [issue](https://github.com/pointfreeco/swift-case-paths/issues/63) we encountered.


rastersize (approved):
Thanks!

PR #185, opened 2022-01-12
Fix copyright headers

kmcbride:
These files have, incorrectly, been using the copyright notice for Apache Software Foundation projects rather than [the one recommended in the license](https://github.com/spotify/Mobius.swift/blob/master/LICENSE#L189-L201).


rastersize (approved):

rastersize:
Good catch!

PR #186, opened 2022-01-24
Consolidate into package

kmcbride:
Recreating this PR for a few reasons:
- Simplifies maintenance
  - Compatibility and releases are easier to manage (as illustrated by our oft-outdated CocoaPods specs)
- Removes excess infrastructure
  - Projects, Schemes, and Configs are no longer needed
- Improves the development workflow
  - Building and testing for platforms besides iOS previously needed bootstrap script and project configuration adjustments
- Helps with other tooling
  - Declaring CasePaths as an explicit dependency would be overly cumbersome if also continuing to support CocoaPods, but keeping it external is preferable because it allows us to more easily integrate and attribute it via our tooling


rastersize (approved):
This makes sense to me.

I do wonder how many, if any?, folks use CocoaPods to integrate Mobius? I guess anyone that need it could either create their own spec or fork?

kmcbride:
It's hard to say, but my guess would be very few. I tried searching across GitHub and [practically nothing](https://github.com/search?q=MobiusCore+filename%3APodfile&type=Code&ref=advsearch) comes up. There's also a button in the sidebar of the [CocoaPods page](https://cocoapods.org/pods/MobiusCore) that doesn't show any results for apps using it (but maybe that requires additional configuration?).

To the second question, yes, one should fork and create a Podspec if they need CocoaPods compatibility. This is part of the reason why I referred to [importing CasePaths while supporting CocoaPods](https://github.com/pointfreeco/swift-case-paths/issues/2#issuecomment-820506630) as cumbersome 😬

rastersize:
@kmcbride Sounds like we should go ahead and merge this then 👍

PR #190, opened 2022-04-15
Dispose viewConnectable synchronously

kmcbride:
Asynchronously disposing the controller view connection can lead to unexpected behavior (and ultimately crashes) in certain scenarios. It also differs from how other types (effect handler, event source, loop) are disposed: synchronously on the loop queue.

This contrived example demonstrates how the current implementation will lead to a crash:
```swift
// scenario: executing on main queue, ConnectableClass previously attached via .connectView()
loopController.start()

// queues work for next loop pass
DispatchQueue.main.async {
    // crashes in ConnectableClass .connect() because it executes before the queued dispose
    loopController.start()
}

// stops loop but queues ConnectableClass connection dispose on the acceptQueue (main)
loopController.stop()
```

As I wasn't involved in the original design it's possible I'm missing some important context, but I've attempted to annotate my reasoning for the changes as review comments.

### DIFF
--- a/MobiusCore/Source/AsyncDispatchQueueConnectable.swift
+++ b/MobiusCore/Source/AsyncDispatchQueueConnectable.swift
@@ -14,23 +14,14 @@
 
 import Foundation
 
-/// A connectable adapter which imposes asynchronous dispatch blocks around calls to `accept` and `dispose`.
+/// A connectable adapter which imposes asynchronous dispatch blocks around calls to `accept`.
 ///
-/// Creates `Connection`s that forward invocations to `accept` and `dispose` to a connection returned by the underlying
-/// connectable, first switching to the provided `acceptQueue`. In other words, the real `accept` and `dispose` methods
-/// will always be executed asynchronously on the provided queue.
-///
-/// If the connection’s consumer is invoked between the Connectable’s `dispose` and the underlying asynchronous
-/// `dispose`, the call will not be forwarded.
+/// Creates `Connection`s that forward invocations to `accept` to a connection returned by the underlying connectable,
+/// first switching to the provided `acceptQueue`. In other words, the real `accept` method will always be executed
+/// asynchronously on the provided queue.
 final class AsyncDispatchQueueConnectable<Input, Output>: Connectable {
     private let underlyingConnectable: AnyConnectable<Input, Output>
     private let acceptQueue: DispatchQueue
-
-    private enum DisposalStatus: Equatable {
-        case notDisposed
-        case pendingDispose
-        case disposed
-    }
 
     init(
         _ underlyingConnectable: AnyConnectable<Input, Output>,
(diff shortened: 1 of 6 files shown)


MobiusCore/Source/AsyncDispatchQueueConnectable.swift
```diff
@@ -47,21 +38,14 @@ final class AsyncDispatchQueueConnectable<Input, Output>: Connectable {
         self.init(AnyConnectable(underlyingConnectable), acceptQueue: acceptQueue)
     }
 
-    func connect(_ consumer: @escaping (Output) -> Void) -> Connection<Input> {
-        let disposalStatus = Synchronized(value: DisposalStatus.notDisposed)
+    func connect(_ consumer: @escaping Consumer<Output>) -> Connection<Input> {
+        let protectedConsumer = Synchronized<Optional<Consumer<Output>>>(value: consumer)
```
kmcbride, line 42: Synchronizing on the consumer allows for both tracking the disposal status and clearing the reference when it is no longer usable.

MobiusCore/Source/AsyncDispatchQueueConnectable.swift
```diff
@@ -47,21 +38,14 @@ final class AsyncDispatchQueueConnectable<Input, Output>: Connectable {
...
-            // underlying consumer must not call our connection’s acceptClosure or disposeClosure, or it will deadlock.
-            //
-            // This is OK in our existing use case because the underlying consumer is always flipEventsToLoopQueue from
-            // MobiusController’s initializer, which enters the actual Mobius loop asynchronously. If we exposed this
-            // class, it would be a scary edge case.
-            disposalStatus.read { status in
-                guard status != .pendingDispose else { return }
+            protectedConsumer.read { consumer in
+                guard let consumer = consumer else {
+                    MobiusHooks.errorHandler("cannot consume value after dispose", #file, #line)
```
kmcbride, line 47: The previous code had an issue with retaining the consumer reference, which would lead to a somewhat nebulous crash when its unowned loop reference was accessed by emitting an event after disposal. This code is functionally equivalent, but the cause of the error should be clearer now.

MobiusCore/Source/MobiusController.swift
```diff
@@ -94,18 +94,13 @@ public final class MobiusController<Model, Event, Effect> {
...
         // The input will be the core `MobiusLoop`’s event dispatcher, which asserts that it isn’t invoked after the
-        // loop is disposed. This doesn’t play nicely with asynchrony, so here we assert when the transformed event
-        // consumer is invoked, but fail silently if the controller is stopped before the asynchronous block executes.
+        // loop is disposed. This doesn’t play nicely with asynchrony, so here we fail silently if the controller is
+        // stopped before the asynchronous block executes.
         func flipEventsToLoopQueue(consumer: @escaping Consumer<Event>) -> Consumer<Event> {
             return { event in
-                guard state.running else {
-                    MobiusHooks.errorHandler("\(Self.debugTag): cannot accept events when stopped", #file, #line)
-                }
```
kmcbride, line 107: Due to this closure being created and owned by the loop—and the lifetime of the loop only being tied to the controller's running state—this doesn't appear to be a valid scenario, unless the loop reference was somehow being leaked.

MobiusCore/Source/MobiusController.swift
```diff
@@ -211,9 +206,11 @@ public final class MobiusController<Model, Event, Effect> {
                 var disposables: [Disposable] = [loop]
 
                 if let viewConnectable = stoppedState.viewConnectable {
-                    let viewConnection = viewConnectable.connect { [unowned loop] event in
-                        // Note: loop.unguardedDispatchEvent will call our flipEventsToLoopQueue, which implements the
-                        //       assertion “unguarded” refers to, and also (of course) flips to the loop queue.
+                    let viewConnection = viewConnectable.connect { [weak loop] event in
+                        guard let loop = loop else {
+                            MobiusHooks.errorHandler("\(Self.debugTag): cannot use invalid consumer", #file, #line)
```
kmcbride, line 211: This scenario is already prevented by the `AsyncDispatchQueueConnectable`, but I've added it here just in case the `viewConnectable` type ever changes and doesn't account for it.

MobiusCore/Source/AsyncDispatchQueueConnectable.swift
erikprice: Maybe just bikeshedding but this might make for a good implementation comment, especially considering how difficult it can be to read the nested parameterized types.

MobiusCore/Source/MobiusController.swift
erikprice: Another explanation that might be helpful to live alongside the code.

rastersize (approved):
This makes sense to me. Might still be good to get some of the original Mobius.swift folks to look it over as you mentioned.

kmcbride:
@JensAyton would you mind taking a look? I think you probably have the most insight into the existing implementation.

PR #194, opened 2022-05-31
Route enum cases without associated values

kmcbride:
This resolves a small annoyance with effect enums having to conform to `Equatable` in order to use the `.routeCase(…)` API.

This change is backwards compatible and doesn't affect existing implementations.


rastersize (approved):
That is awesome

PR #205, opened 2024-02-12
Add `asDisposable` convenience property to `Task`

louisdebaere:
Offers a convenient way to return Tasks as a `Disposable` instead of having to wrap it with `AnonymousDisposable`, for example in an `EffectHandler`

**Example usage**
```
extension LoadLyricsModelEffectHandler: EffectHandler {
    func handle(_ playbackData: PlaybackData, _ callback: EffectCallback<CardViewEvent>) -> Disposable {
        Task {
            do {
                let model = try await dataLoader.fetch(for: playbackData.track)
                callback.send(.receivedLyricsModel(model, for: playbackData))
            } catch {
                callback.send(.receivedError(for: playbackData))
            }
        }
        .asDisposable
    }
}
```

### DIFF
--- a/MobiusCore/Source/Disposables/Task+Disposable.swift
+++ b/MobiusCore/Source/Disposables/Task+Disposable.swift
@@ -1,6 +1,28 @@
+// Copyright 2019-2024 Spotify AB.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
 public extension Task {
 
+    /// A disposable for use with `EffectHandler` that will cancel the task
+    ///
+    ///     func handle(_ parameters: Void, _ callback: EffectCallback<Event>) -> Disposable {
+    ///         Task {
+    ///
+    ///         }
+    ///        .asDisposable
+    ///     }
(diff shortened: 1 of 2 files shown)


MobiusCore/Source/Disposables/Task+Disposable.swift
```diff
@@ -0,0 +1,7 @@
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+public extension Task {
+
+    var asDisposable: any Disposable {
```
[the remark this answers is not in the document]
louisdebaere: How's this?

MobiusCore/Source/Disposables/Task+Disposable.swift
```diff
@@ -0,0 +1,29 @@
...
+
+    /// A disposable for use with `EffectHandler` that will cancel the task
+    ///
+    ///     func handle(_ parameters: Void, _ callback: EffectCallback<Event>) -> Disposable {
+    ///         Task {
+    ///
+    ///         }
+    ///        .asDisposable
+    ///     }
+    var asDisposable: any Disposable {
```
zvonicek, line 26: Existential types are more expensive than concrete types, so it's better to avoid them if possible. In this case it should be easily avoidable by using `some` instead of `any`.

zvonicek (commented):
I noticed one more thing that's worth changing

MobiusCore/Source/Disposables/Task+Disposable.swift
louisdebaere: Cool, when compiled, that's the same thing as `var asDisposable: AnonymousDisposable`, correct?

zvonicek: Exactly, you could therefore also return just `AnonymousDisposable`. But I think that here the fact that the `Disposable` is of type `AnonymousDisposable` is insignificant from the API perspective, so it might be better to use the opaque return type (`some`).

louisdebaere: Agreed

louisdebaere:
Pinging

zvonicek (approved):

PR #209, opened 2024-02-29
Add EmptyDisposable

mjburghard:


MobiusCore/Source/Disposables/EmptyDisposable.swift
```diff
@@ -0,0 +1,30 @@
...
+    /// Create an `EmptyDisposable`
+    public init() {}
+
+    public func dispose() {
+        // No-op
+    }
+}
+
+public extension Disposable where Self == EmptyDisposable {
+    static var empty: EmptyDisposable { EmptyDisposable() }
```
zvonicek, line 29: Since the disposable will be most of the times used through this getter, I would suggest to add documentation to it, such as:

```suggestion
    /// Gets the disposable that does nothing when disposed.
    static var empty: EmptyDisposable { EmptyDisposable() }
```

zvonicek, line 29: Since we don't need the concrete type outside, could you return just `Disposable`? (on iOS 13+ even `some Disposable`, but I think that the target here is lower)

zvonicek (commented):
overall looks like a good convenience to add

PR #210, opened 2024-04-23
Deprecate the EffectHandler protocol

kmcbride:
I propose deprecating `EffectHandler` and removing it in a future release because I believe it leads to frequent misuse of the API without providing any obvious advantages. This removal will ideally encourage users to be more intentional about which type of handler they route to.

### DIFF
--- a/MobiusCore/Source/EffectHandlers/EffectExecutor.swift
+++ b/MobiusCore/Source/EffectHandlers/EffectExecutor.swift
@@ -4,7 +4,13 @@
 import Foundation
 
 final class EffectExecutor<Effect, Event>: Connectable {
-    private let handleEffect: (Effect, EffectCallback<Event>) -> Disposable
+    enum Operation {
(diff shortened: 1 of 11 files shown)


MobiusCore/Source/EffectHandlers/EffectExecutor.swift
```diff
@@ -15,19 +15,29 @@
...
+    private let operation: Operation
     private var output: Consumer<Event>?
 
     private let lock = Lock()
 
     // Keep track of each received effect's state.
     // When an effect has completed, it should be removed from this dictionary.
     // When disposing this effect handler, all entries must be removed.
-    private var handlingEffects: [Int64: EffectHandlingState<Event>] = [:]
+    private var ongoingEffects: [Int64: EffectHandlingState<Event>] = [:]
```
erikprice, line 32: 💯 this rename

erikprice (approved):
Some great refactoring in this PR. I think I agree with the deprecation as well.

PR #217, opened 2025-08-21
Upgrade Nimble (and min SDK versions)

dflems:
`Nimble.Predicate` has become `Nimble.Matcher`. Other than that, Nimble seems happy.

Also bumps the min SDK version to match latest Nimble.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.2
+// swift-tools-version:5.7
 import PackageDescription
 
 let depCasePaths: PackageDescription.Target.Dependency = .product(name: "CasePaths", package: "swift-case-paths")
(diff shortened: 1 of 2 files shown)


Package.swift
```diff
@@ -8,10 +8,10 @@ let depNimble: PackageDescription.Target.Dependency = .product(name: "Nimble", p
...
     name: "Mobius",
     platforms: [
-        .iOS(.v10),
-        .macOS(.v10_12),
-        .tvOS(.v10),
-        .watchOS(.v3),
+        .iOS(.v13),
+        .macOS(.v10_15),
+        .tvOS(.v13),
+        .watchOS(.v6),
```
dflems, line 14: Nimble complained about some of these. Probably about time to bump anyway? iOS 10 is mighty old.

Package.swift
```diff
@@ -21,7 +21,7 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/pointfreeco/swift-case-paths", .upToNextMinor(from: "0.10.1")),
-        .package(url: "https://github.com/Quick/Nimble", from: "10.0.0"),
+        .package(url: "https://github.com/Quick/Nimble", from: "11.0.0"),
```
dflems, line 24: Will followup w/ an upgrade to 13.x, this is a baby step

zvonicek (approved):

PR #218, opened 2026-03-04
Add Swift 5.10+ onMainActor route builder

louisdebaere:
## Why make this change

Mobius supports routing effects to the main thread via `.on(queue: .main)`, but when the handler touches `@MainActor`-isolated state, Swift requires the call site to manually bridge isolation — typically with `MainActor.assumeIsolated`. This scatters a framework-level concern across consumer code and creates repetitive boilerplate that's easy to get wrong.

As projects adopt Swift Concurrency and `@MainActor` annotations spread (especially with Swift 6.2's `defaultIsolation(MainActor.self)`), this friction multiplies — every main-queue effect route needs the same wrapper.

This change introduces a Swift Concurrency-aware route that centralises the isolation bridge inside Mobius, keeping consumer code clean while preserving the existing dispatch semantics.

## What will change

- Add `_PartialEffectRouter.onMainActor()` returning a dedicated `_MainActorPartialEffectRouter`.
- Add `.to(...)` overloads on `_MainActorPartialEffectRouter` accepting `@MainActor @Sendable` closures:
  - `(EffectParameters) -> Void`
  - `() -> Void` (when `EffectParameters == Void`)
- Internally delegate to the existing `.on(queue: .main)` route path, then invoke closures via `MainActor.assumeIsolated` — no new dispatch mechanism, no `Task`, no ordering changes.
- Gate the API and related tests behind `#if compiler(>=5.10)`.

## Before / After

Manual bridge at every call site:
```swift
.routeEffects(equalTo: .someEffect)
    .on(queue: .main)
    .to {
        MainActor.assumeIsolated {
            presenter.show()
        }
    }

.routeEffects(withParameters: extract)
    .on(queue: .main)
    .to { params in
        MainActor.assumeIsolated {
            handler.handle(params)
        }
    }
```

Bridge handled by the framework:
```swift
.routeEffects(equalTo: .someEffect)
    .onMainActor()
    .to {
        presenter.show()
    }

.routeEffects(withParameters: extract)
    .onMainActor()
    .to { params in
        handler.handle(params)
    }
```

Existing queue-based routes are completely unaffected:
```swift
.routeEffects(equalTo: .otherEffect)
    .on(queue: backgroundQueue)
    .to {
        processor.process()
    }
```

On compilers older than 5.10, `.onMainActor()` is unavailable by design. The existing `.on(queue: .main).to { MainActor.assumeIsolated { … } }` pattern remains the compatibility path.

## Why this shape

**Queue dispatch and actor isolation are separate concepts.** `.on(queue:)` is a GCD concern; `.onMainActor()` is a Swift Concurrency concern. Giving them distinct API surfaces makes intent clear and avoids conflating the two models.

**No new dispatch path.** `.onMainActor()` delegates to the same `.on(queue: .main)` routing that already exists. Dispatch order, callback lifecycle, and `Disposable` semantics are identical — the only addition is the `assumeIsolated` bridge.

**Centralised bridge.** `MainActor.assumeIsolated` moves from potentially hundreds of call sites into one place in the framework, where it's structurally correct (the main queue is the MainActor executor) and documented.

**Ready for Swift 6.2 consumer settings.** When consumers enable `defaultIsolation(MainActor.self)` or Approachable Concurrency (`NonisolatedNonsendingByDefault`), their closures are implicitly `@MainActor` — the compiler picks the `.onMainActor().to { }` overload without any annotation at the call site. Without these overloads, those same consumers would face type mismatches or be forced to add `nonisolated` workarounds to use the existing queue-based API.

## Alternatives considered

- **`@MainActor` overloads on the queue-agnostic builder.** Simpler, but permits `.on(queue: .global()).to(@MainActor ...)` — shifts misconfiguration from compile time to a runtime trap.
- **`Task { @MainActor in }` instead of `assumeIsolated`.** Safe by construction, but introduces an async hop on top of the queue dispatch, changing ordering guarantees and complicating `callback.end()` placement.
- **Typed main-queue token via `@_disfavoredOverload`.** Compile-time enforcement, but relies on underscored API, breaks when aliasing `DispatchQueue.main`, and would strand `_PartialEffectRouter` extensions that return queue-bound builders.
- **Keep current call-site pattern only.** No framework change, but leaves repetitive and easy-to-miss bridging in consumer code — increasingly painful as `@MainActor` adoption grows.

Given these trade-offs, `.onMainActor().to { }` is the minimal, explicit, and forwards-compatible addition.


MobiusCore/Source/EffectHandlers/EffectRouterDSL.swift
```diff
@@ -35,6 +37,16 @@ public extension _PartialEffectRouter {
         }
     }
 
+    /// Route main-isolated effects through the same queue path as `.on(queue: .main)`.
+    ///
+    /// This returns a dedicated builder exposing `to(...)` for `@MainActor` closures.
+    #if compiler(>=5.10)
+    @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.0, *)
+    func onMainActor() -> _MainActorPartialEffectRouter<Effect, EffectParameters, Event> {
+        return _MainActorPartialEffectRouter(partialRouter: on(queue: .main))
```
kmcbride: nitpick: might be better to enforce the queue switch within a `_MainActorPartialEffectRouter` initializer

```swift
init(partialRouter: _PartialEffectRouter<Effect, EffectParameters, Event>) {
    self.partialRouter = partialRouter.on(queue: .main)
}
```

kmcbride (dismissed):

MobiusCore/Source/EffectHandlers/EffectRouterDSL.swift
```diff
@@ -35,6 +37,16 @@ public extension _PartialEffectRouter {
         }
     }
 
+    /// Route main-isolated effects through the same queue path as `.on(queue: .main)`.
+    ///
+    /// This returns a dedicated builder exposing `to(...)` for `@MainActor` closures.
+    #if compiler(>=5.10)
+    @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.0, *)
+    func onMainActor() -> _MainActorPartialEffectRouter<Effect, EffectParameters, Event> {
```
zvonicek: I'd prefer if this can be `var` instead of `func`, which I believe would be more idiomatic for cases like this in Swift (e.g. [var main on DispatchQueue](https://developer.apple.com/documentation/dispatch/dispatchqueue/main))

MobiusCore/Source/EffectHandlers/EffectRouterDSL.swift
```diff
@@ -51,3 +63,48 @@ public extension _PartialEffectRouter {
...
+
+#if compiler(>=5.10)
+/// A `_MainActorPartialEffectRouter` represents the state between an `onMainActor` call and a `to`.
+///
+/// Client code should not refer to this type directly.
+public struct _MainActorPartialEffectRouter<Effect, EffectParameters, Event> {
+    fileprivate let partialRouter: _PartialEffectRouter<Effect, EffectParameters, Event>
+}
+
+public extension _MainActorPartialEffectRouter {
```
zvonicek, line 79: Can we also add support for non-fire and forget variants, such as 

```swift
    public func to(
        _ handle: @escaping @MainActor (EffectParameters, EffectCallback<Event>) -> any Disposable
    ) -> EffectRouter<Effect, Event> {
```

and 

```swift
    public func toEvent(
        _ eventClosure: @escaping @MainActor (EffectParameters) -> Event?
    ) -> EffectRouter<Effect, Event> {
```

zvonicek (approved):

